PR #41, opened 2022-08-05
Feat/make dapp nextjs

devceline:
# Changes
- Move `react-dapp-v2` from a SPA to a nextjs app


dapps/react-dapp-v2/package.json
```diff
@@ -1,31 +1,13 @@
...
-  "bugs": {
-    "url": "https://github.com/walletconnect/walletconnect-monorepo/issues"
-  },
-  "resolutions": {
-    "react-error-overlay": "6.0.9"
+    "dev": "next dev",
+    "build": "next build",
+    "start": "next start",
+    "lint": "next lint",
+		"prettier:write": "prettier --write '**/*.{js,ts,jsx,tsx}'"
```
bkrem, line 10: Indentation

dapps/react-dapp-v2/src/pages/api/hello.ts
```diff
@@ -0,0 +1,13 @@
+// Next.js API route support: https://nextjs.org/docs/api-routes/introduction
```
bkrem, line 1: Can remove this since it's just the boilerplate API route.

bkrem (changes requested):
One thing I can't comment on the file directly in the diff: I think we can remove https://github.com/WalletConnect/web-examples/pull/41/files#diff-fbc553e80c2f002d9c416efaec6a1364dfb23ff4b753156d986e0e8611a909f2 (leftover `react-app-env.d.ts`)

Great stuff and I still can't keep up with your turnaround time! 😄 A few small things I think we need to tweak, but I gave it a spin both on the deploy and locally and it's looking really good 🙌 

Will wait with merging upstream to `main` until Monday regardless, it seems totally fine but I'd rather leave the tried-and-true version up for the weekend to avoid possibly blocking people from testing v2 👍

bkrem (approved):

PR #47, opened 2022-08-25
Fix incorrect env var ref

devceline:
# Changes

- Change `process.env.RELAY_URL` to `process.env.NEXT_PUBLIC_RELAY_URL`


IljaDaderko (approved):
👍

PR #51, opened 2022-09-02
Remove session from nav for auth

devceline:
# Changes

- Removed sessions from nav as the concept doesn't exist in auth

### DIFF
--- a/dapps/react-dapp-auth/pages/index.tsx
+++ b/dapps/react-dapp-auth/pages/index.tsx
@@ -13,7 +13,7 @@ import AuthClient from "@walletconnect/auth-client";
 import type { NextPage } from "next";
 import Link from "next/link";
 import Qrcode from "qrcode";
-import { useCallback, useEffect, useRef, useState } from "react";
+import { Fragment, useCallback, useEffect, useRef, useState } from "react";
 
 const Home: NextPage = () => {
   const [client, setClient] = useState<AuthClient | null>();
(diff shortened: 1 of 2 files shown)


dapps/react-dapp-auth/pages/index.tsx
```diff
@@ -91,18 +91,26 @@ const Home: NextPage = () => {
...
-                  ref={canvasRef}
-                />
+                    }}
+                    ref={canvasRef}
+                  />
+                  <Text
+                    color="gray.400"
+                    style={{ fontStyle: "italic", fontSize: "0.75em" }}
+                  >
+                    (You can copy the URI by clicking the QR code above)
```
bkrem, line 111: Nice 👍

bkrem:
Ah the central CTA is of course off centre now because we have even amounts of nav items instead of odd now.

I think probably best to to just leave a spacer element there or give the chains list icon double spacing until we have some useful replacement 👍

709126F0-0615-4944-AA4A-77A4B97E6F7E

devceline:
> Ah the central CTA is of course off centre now because we have even amounts of nav items instead of odd now.
> 
> I think probably best to to just leave a spacer element there or give the chains list icon double spacing until we have some useful replacement +1
> 
> 709126F0-0615-4944-AA4A-77A4B97E6F7E

I

> Ah the central CTA is of course off centre now because we have even amounts of nav items instead of odd now.
> 
> I think probably best to to just leave a spacer element there or give the chains list icon double spacing until we have some useful replacement +1
> 
> 709126F0-0615-4944-AA4A-77A4B97E6F7E

Looks weird with a spacer or giving the chains 2 spaces and centering that
image

bkrem:
> Looks weird with a spacer or giving the chains 2 spaces and centering that

Also true. @IljaDaderko any quick solutions you're aware of? Not worth bikeshedding any more on this detail in my opinion though so let's just go for it for now if not 👍

bkrem (approved):

IljaDaderko:
> > Looks weird with a spacer or giving the chains 2 spaces and centering that
> 
> Also true. @IljaDaderko any quick solutions you're aware of? Not worth bikeshedding any more on this detail in my opinion though so let's just go for it for now if not 👍

Mmm, I guess we can move WC button to the right and rest of the menu to the left, so it doesn't require even amount of items on each side. Kinda like in google's material design with that "fab" button

Issue #57, opened 2022-09-09
Allow to select the region


arein:
We are regionalizing. By default the web-examples should connect to `relay.walletconnect.com` or `staging.relay.walletconnect.com` for lower environments.

In order to properly test x-region scenarios we need to select the regionalized endpoint.

So add a way to select between:
* `us-east-1.relay.walletconnect.com`
* `eu-central-1.relay.walletconnect.com`
* `ap-southeast-east-1.relay.walletconnect.com`
or for staging:
* `us-east-1.staging.relay.walletconnect.com`
* `eu-central-1.staging.relay.walletconnect.com`
* `ap-southeast-east-1.staging.relay.walletconnect.com`

See https://github.com/WalletConnect/rs-relay/issues/302 for more info

PR #58, opened 2022-09-12
Feat/region relayer dropdown

crypblizz8:
**Overview**
- Worked on adding a DropDown menu for regions and selecting different relayers
- As per #57 
- [Loom Recording](https://www.loom.com/share/f22c2ccd23ff4dcc92238a61f34b6e03)

**_Note_**
As the new relayers go live (~ Wednesday), they should function to resolve the new createClient() result when live. Hence the Draft PR.

**Process**
- Understanding `ClientContext.tsx `&` Index.tsx`
- Added `regionRelayer` and `setRegionRelayer` state into the UseMemo + Interface
- Add a condition if the Relayer changes from the `DEFAULT_RELAY_URL`, a new `createClient()` will be instantiated
- Work on a new Component called `Dropdown.tsx`
- Used React-Select to make it easier to not build out from scratch. If it is preferred to not use libraries and more vanilla, I can re-create it.

**How to Reproduce**
- Use this branch + do a clean `npm i`
- Add the standard Project id details as outlined in the ReadME file
- `Yarn dev`
- Check the console logs for the correct RELAYERS have been instantiated.


arein (approved):
LGTM! Thank you! :)

dapps/react-dapp-v2/src/constants/default.ts
```diff
@@ -25,7 +25,7 @@ export const DEFAULT_TEST_CHAINS = [
 export const DEFAULT_CHAINS = [...DEFAULT_MAIN_CHAINS, ...DEFAULT_TEST_CHAINS];
 
 export const DEFAULT_PROJECT_ID = process.env.NEXT_PUBLIC_PROJECT_ID;
-export const DEFAULT_RELAY_URL = process.env.NEXT_PUBLIC_RELAY_URL;
+export const DEFAULT_RELAY_URL = process.env.NEXT_PUBLIC_RELAY_URL; // set a conditional here ...
```
bkrem, line 28: Is the comment still relevant? If so could you expand a bit on what your thinking was here so I get it :)

dapps/react-dapp-v2/src/contexts/ClientContext.tsx
```diff
@@ -230,15 +236,16 @@ export function ClientContextProvider({
...
         logger: DEFAULT_LOGGER,
-        relayUrl: DEFAULT_RELAY_URL,
+        relayUrl: relayerRegion, // How do we change here // Shoudl be relayerRegion
         projectId: DEFAULT_PROJECT_ID,
         metadata: getAppMetadata() || DEFAULT_APP_METADATA,
       });
 
       console.log("CREATED CLIENT: ", _client);
+      console.log("relayerRegion ", relayerRegion);
+      // console.log("CLIENT RELAYER: ", _client?.opts?.relayUrl);
```
bkrem, line 248: Unless definitely needed at a later point:

```suggestion

```

bkrem (approved):
Looking good, thanks for tackling this on day 1! 🙌  Also loving that you included a demo vid for a small feature like this 💯 

The preview build for `react-dapp-v2` seems to be failing on a TS issue (see [here](https://vercel.com/walletconnect/react-dapp-v2/4hrES9kkpKXyvpR6cdzGXf5NDLsd), or in the Vercel bot comment above).

Once that's resolved we should be all good to go 🚢

dapps/react-dapp-v2/src/constants/default.ts
crypblizz8: Removed~!

dapps/react-dapp-v2/src/contexts/ClientContext.tsx
crypblizz8: Removed~!

crypblizz8:
Rewrote in Select Component instead of using a library

[Loom Video of the latest:  ](https://www.loom.com/share/5e0dfe988ad14f02ba4b6f4b4f078bf9)

AsiaPacific seems to still not be running though.

chadyj:
Passing on the UX feedback from Slack.

Since this web example is public facing, and this region feature is an internal feature could we change the UI to cater to the main audience better?  Also, on mobile the drop down pushes the main functionality further down, and implies a selection needs to be made.

Screenshot 2022-09-14 at 09 56 05


A simple suggestion is move it to the bottom.  Or move to the bottom, and hide it in a "Options" icon or something similar. In most cases this setting never needs to be changed, so it can be hidden/de-ephemphasied in the UI. 

Also for the staging version, are staging relays being used?

crypblizz8:
@chadyj great suggestion, have amended + added to the bottom in a settings button. If more styles is required, let me know! CC @arein 

[Loom Video](https://www.loom.com/share/a8213e6782b442d3ad4e7297287d76fa)

chadyj:
Looks groovy thanks @crypblizz8

dapps/react-dapp-v2/src/contexts/ClientContext.tsx
```diff
@@ -70,10 +70,16 @@ export function ClientContextProvider({
   const [solanaPublicKeys, setSolanaPublicKeys] =
     useState<Record<string, PublicKey>>();
   const [chains, setChains] = useState<string[]>([]);
+  // const [relayerRegion, setRelayerRegion] = useState<string>(DEFAULT_RELAY_URL);
```
arein: cruft?

arein (approved):

PR #71, opened 2022-11-01
fix(wallet): NEXT_PUBLIC_RELAY_URL reference

hanwong:
fix name NEXT_PUBLIC_RELAY_URL


bkrem (approved):
Great catch, thank you! 💯

PR #77, opened 2022-11-14
refactor(polkadot): replace scale encoded tx with JSON payload

hamidra:
The previous Polkadot example was using a hex representation of the [SCALE](https://docs.substrate.io/reference/scale-codec/) encoded payload of the transaction. 
These changes will add the JSON payload validation to the wallet and uses a JSON payload to submit Polkadot transactions, which is a closer representation of what actually happens in the wallets and dapps.


wallets/react-wallet-v2/src/utils/PolkadotWalletUtil.ts
```diff
@@ -11,6 +12,14 @@ let address2: string
 /**
  * Utilities
  */
+export function getPolkadotWallets(address: string) {
+  Object.entries(polkadotWallets)
```
bkrem, line 16: This expression isn't assigned/seems to duplicate the line below

wallets/react-wallet-v2/src/utils/PolkadotRequestHandlerUtil.ts
```diff
@@ -10,15 +10,20 @@ export async function approvePolkadotRequest(
 ) {
   const { params, id } = requestEvent
   const { request } = params
-  const wallet = polkadotWallets[getWalletAddressFromParams(polkadotAddresses, params)]
+  const address = request.params?.address
+  const wallet = getPolkadotWallets(address)
+
+  if (!wallet) {
+    throw new Error('Wallet does not exist')
```
bkrem, line 17:
```suggestion
    throw new Error('Polkadot wallet does not exist')
```

wallets/react-wallet-v2/src/utils/PolkadotRequestHandlerUtil.ts
```diff
@@ -10,15 +10,20 @@ export async function approvePolkadotRequest(
...
+  }
 
   switch (request.method) {
     case POLKADOT_SIGNING_METHODS.POLKADOT_SIGN_MESSAGE:
-      const signature = await wallet.signMessage(request.params.message)
+      const signature = await wallet?.signMessage(request.params.message)
```
bkrem, line 22: Is the optional chaining needed with the reference guard (`if (!wallet)`) above?

dapps/react-dapp-v2/src/contexts/JsonRpcContext.tsx
```diff
@@ -669,19 +688,11 @@ export function JsonRpcContextProvider({
...
-            result.signature,
-            address
-          );
-
+          
           return {
             method: DEFAULT_POLKADOT_METHODS.POLKADOT_SIGN_TRANSACTION,
             address,
-            valid,
+            valid: true,
```
bkrem, line 695: This seems like a step backwards if we're hardcoding `valid`. Is there a reason we can't use dynamic signature validation as it was?

dapps/react-dapp-v2/src/contexts/JsonRpcContext.tsx
```diff
@@ -642,18 +641,38 @@ export function JsonRpcContextProvider({
...
     testSignTransaction: _createJsonRpcRequestHandler(
       async (
         chainId: string,
         address: string
       ): Promise<IFormattedRpcResponse> => {
-        // Below example is a scale encoded payload for system.remark("this is a test wallet-connect remark") transaction.
-        // decode url: https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Frpc.polkadot.io#/extrinsics/decode/0x00019074686973206973206120746573742077616c6c65742d636f6e6e6563742072656d61726b
-        const transactionPayload =
-          "0x00019074686973206973206120746573742077616c6c65742d636f6e6e6563742072656d61726b05010000222400000d00000091b171bb158e2d3848fa23a9f1c25182fb8e20313b2c1eb49219da7a70ce90c3dc1f37ce7899cf20f63f5ea343f33e9e7b229c7e245049c2a7afc236861fc8b4";
+        const transactionPayload = {
```
bkrem, line 652: Nice 🙌  thanks for adding a more human-friendly payload 😄

wallets/react-wallet-v2/src/utils/PolkadotWalletUtil.ts
hamidra: Uh. good catch!

wallets/react-wallet-v2/src/utils/PolkadotRequestHandlerUtil.ts
hamidra: right. can be removed.

dapps/react-dapp-v2/src/contexts/JsonRpcContext.tsx
hamidra: I did not want to add a dependency to Polkadot RPCs for this validation. Polkadot.js apis currently uses the RPC for validating a tx + signature but there might be still some way to just validate the signature which I need to do some research to find out. I'll send a separate PR when I find the offline signature validation.

dapps/react-dapp-v2/src/contexts/JsonRpcContext.tsx
hamidra: I did not want to add a dependency to Polkadot RPCs for this validation. Polkadot.js apis currently uses the RPC for validating a tx + signature but there might be still some way to just validate the signature which I need to do some research to find out. I'll send a separate PR when I find the offline signature validation.

dapps/react-dapp-v2/src/contexts/JsonRpcContext.tsx
hamidra: Thanks. Hopefully, this makes it easier to read the transactions, and also helps with testing 🙌

dapps/react-dapp-v2/src/contexts/JsonRpcContext.tsx
bkrem: Got it, thanks for the context 👍

bkrem (approved):

bkrem:
Thanks again for improving the polkadot reference! 🙏

PR #79, opened 2022-11-16
Use standalone modal

devceline:
# Changes
- Remove `QrView` and `qrcode` package
- Add `@web3modal/ethereum`, `@web3modal/ui` and `@web3modal/core` and use standalone modal when authenticating


IljaDaderko (commented):
Looks like explorer requests in the preview are failing. Seems they are missing projectId, maybe env var missing or something?

dapps/react-dapp-auth/package.json
```diff
@@ -13,25 +13,27 @@
     "@emotion/react": "^11.10.0",
     "@emotion/styled": "^11.10.0",
     "@walletconnect/auth-client": "2.0.4",
+    "@web3modal/core": "^2.0.0-beta.4",
```
bkrem, line 16: Lockfile diff is too big to open for me here but I suspect these are technically also at `2.2.2` in line with `standalone` below?

Could we use the explicit versions for these too instead of the stale beta prereleases to avoid possible confusion?

dapps/react-dapp-auth/pages/index.tsx
```diff
@@ -1,13 +1,32 @@
...
 import type { NextPage } from "next";
 import { useCallback, useEffect, useState } from "react";
 import DefaultView from "../views/DefaultView";
-import QrView from "../views/QrView";
 import SignedInView from "../views/SignedInView";
+import type { W3mModal } from "@web3modal/ui";
+import "@web3modal/ui";
 
-console.log(`AuthClient@${version}`);
+declare global {
```
bkrem, line 11: Think this ceremony and the explicit `w3m-modal` web component have been factored out in the meantime:

https://docs.walletconnect.com/2.0/web3modal/standalone/installation

Will defer to @0xAsimetriq here though on the details

bkrem (approved):
Thank you! 🙏

PR #86, opened 2022-11-29
Add Tron Chain to the Dapp example and wallet

myzykj:
Add Tron Chain to the Dapp example([react-dapp-v2](https://github.com/WalletConnect/web-examples/tree/main/dapps/react-dapp-v2)) and wallet([react-wallet-v2](https://github.com/WalletConnect/web-examples/tree/main/wallets/react-wallet-v2))


dapps/react-dapp-v2/src/contexts/JsonRpcContext.tsx
```diff
@@ -18,6 +18,8 @@ import {
   SystemProgram,
   Transaction as SolanaTransaction,
 } from "@solana/web3.js";
+// @ts-ignore
```
bkrem, line 21: Why is the `@ts-ignore` directive needed here?

If needed, please use `@ts-expect-error - [explanation]` so there's context and this can later be removed if no longer a TS error since `expect-error` will then flag it again.

dapps/react-dapp-v2/src/contexts/JsonRpcContext.tsx
```diff
@@ -1035,6 +1042,85 @@ export function JsonRpcContextProvider({
...
   };
 
+   // -------- TRON RPC METHODS --------
+
+  const tronRpc = {
+    testSignTransaction: _createJsonRpcRequestHandler(
+      async (chainId: string, address: string): Promise<IFormattedRpcResponse> => {
+        // Nile TestNet, if you want to use in MainNet, change the fullHost to 'https://api.trongrid.io'
+        const tronWeb = new TronWeb({
+          fullHost: 'https://nile.trongrid.io/',
```
bkrem, line 1052: If a testnet host is being hardcoded here to generate a test transaction, will this cause failures when calling this against Tron mainnet since the TX was generated for testnet?

dapps/react-dapp-v2/src/contexts/JsonRpcContext.tsx
```diff
@@ -1035,6 +1042,85 @@ export function JsonRpcContextProvider({
...
+          };
+        } catch (error: any) {
+          throw new Error(error);
+        }
+      }
+    ),
+    testSignMessage: _createJsonRpcRequestHandler(
+      async (chainId: string, address: string): Promise<IFormattedRpcResponse> => {
+
+        var message = 'TestTronString';
```
bkrem, line 1095: Code style nit: please use `let` over `var`. Looks like message is never reassigned so suggesting `const`:
```suggestion
        const message = 'This is a message to be signed for Tron';
```

wallets/react-wallet-v2/src/lib/TronLib.ts
```diff
@@ -0,0 +1,50 @@
...
+    this.privateKey = privateKey
+    this.tronWeb = new TronWeb({
+      // Nile TestNet, if you want to use in MainNet, change the fullHost to 'https://api.trongrid.io', or use tronWeb.setFullNode
+      fullHost: 'https://nile.trongrid.io/',
+      privateKey: privateKey
+    })
+  }
+
+  static async init({ privateKey }: IInitArguments) {
+    // wait till  WASM is initialized, in case it is not initialized already (WASM is required for 'sr25519').
```
bkrem, line 28: Accidental leftover comment?

wallets/react-wallet-v2/src/utils/TronWalletUtil.ts
```diff
@@ -0,0 +1,48 @@
...
+    tronWeb2 = await TronLib.init({ privateKey: '' })
+
+    const account1 = await tronWeb1.createAccount()
+    const account2 = await tronWeb2.createAccount()
+    // Don't store privateKey in local storage in a production project!
+    localStorage.setItem('TRON_PrivateKey_1', account1.privateKey)
+    localStorage.setItem('TRON_PrivateKey_2', account2.privateKey)
+  }
+
+  // @ts-ignore
```
bkrem, line 32: Context please on the `@ts-ignore` usage here/can this be avoided?

bkrem (changes requested):
Thanks for putting together this PR to add Tron @myzykj!

Some minor tweaks so far but looking good in general.

Using the example dapp and wallet together seems to work fine, but are you aware of a wallet in the wild that supports TRON that we can verify this against? 
With Solana we saw small but important differences with real world wallets (e.g. using a `UInt8Array` instead of a `base58` string to represent some data).

dapps/react-dapp-v2/src/contexts/JsonRpcContext.tsx
myzykj: use @ts-expect-error now

dapps/react-dapp-v2/src/contexts/JsonRpcContext.tsx
myzykj: check if it is testNet here:

```javascript
const fullHost = isTestnet ? "https://nile.trongrid.io/" : "https://api.trongrid.io/";
const tronWeb = new TronWeb({
          fullHost,
        })
```

dapps/react-dapp-v2/src/contexts/JsonRpcContext.tsx
myzykj: change the code as above

wallets/react-wallet-v2/src/lib/TronLib.ts
myzykj: removed the comment

wallets/react-wallet-v2/src/utils/TronWalletUtil.ts
myzykj: have removed them

dapps/react-dapp-v2/src/contexts/JsonRpcContext.tsx
bkrem: Thanks, still needs an explanation please as to why there is a TS error (i.e. `@ts-expect-error - [explanation]` as shown above)

dapps/react-dapp-v2/src/contexts/JsonRpcContext.tsx
```diff
@@ -1035,6 +1042,85 @@ export function JsonRpcContextProvider({
...
+  const tronRpc = {
+    testSignTransaction: _createJsonRpcRequestHandler(
+      async (chainId: string, address: string): Promise<IFormattedRpcResponse> => {
+        // Nile TestNet, if you want to use in MainNet, change the fullHost to 'https://api.trongrid.io'
+        const tronWeb = new TronWeb({
+          fullHost: 'https://nile.trongrid.io/',
+        })
+
+        const testTransaction = await tronWeb.transactionBuilder.triggerSmartContract(
+          'TXLAQ63Xg1NAzckPwKHvzw7CSEmLMEqcdj',
```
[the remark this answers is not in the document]
bkrem: Thanks 👍

bkrem:
Thanks for addressing the comments and sorry about the wait, was OOO yesterday so only now getting round to re-checking this.

Will do some final checks against the deployments and then should be good to go if everything works as expected 👍

dapps/react-dapp-v2/src/contexts/JsonRpcContext.tsx
myzykj: Because tronWeb doesn't support TypeScript at this moment,  so I add this in case any error

bkrem:
Getting this when trying to sign for Tron testnet:

Screenshot 2022-12-20 at 18 14 49

The Testnet field in the wallet is simply for enabling display of testnets in the list tab, not functionality. This shouldn't be a hard requirement for users to have to enable this first/it's not necessary when signing requests for any other testnet.

myzykj:
Hi @bkrem, 

I have removed the "TestNet enable" check code in the Wallet. Thank you very much.

fa05c8e

bkrem:
Thanks for addressing that @myzykj, signing seems to work as expected now 👍 

Sorry but still finding some weird edge case behaviour in the wallet implementation I didn't see beforehand.

1. Do a fresh init of the wallet (or `localStorage.clear()` in console and reload)
2. Generate URI for Tron testnet on the dapp, paste into wallet
3. Wallet shows the below, i.e. just `false` instead of the actual account selection. Accepting and trying to sign from this state also results in address being set as `false` in the request modals etc.
4. If I hard reload again and retry it works as expected. Presumably a bug in the TRON setup/initialisation.

Screenshot 2022-12-21 at 10 43 14

myzykj:
Hi @bkrem,

I have updated the wallet init function, if there is no privateKey in localStorage, will generate one and return wallet. Before, it return the wallet and generate the account.

I have tested this case now, sorry for the mistake. Thank you very much.

> 72fdf75

PR #91, opened 2022-12-13
fix: eth_signTransaction signer verification for Celo

kathaypacific:
Add custom verification for Celo transaction type when using `eth_signTransaction`. The Celo transaction type does not work with the existing validation method `EthTransaction.fromSerializedTx`.

### DIFF
--- a/dapps/react-dapp-v2/src/contexts/JsonRpcContext.tsx
+++ b/dapps/react-dapp-v2/src/contexts/JsonRpcContext.tsx
@@ -253,10 +253,15 @@ export function JsonRpcContextProvider({
           },
         });
 
+        const CELO_ALFAJORES_CHAIN_ID = 44787;
+        const CELO_MAINNET_CHAIN_ID = 42220;
+
         let valid = false;
         const [, reference] = chainId.split(":");
-        if (reference === "44787" || reference === "42220") {
-          // Verify Celo transaction
+        if (
+          reference === CELO_ALFAJORES_CHAIN_ID.toString() ||
+          reference === CELO_MAINNET_CHAIN_ID.toString()
+        ) {
           const [, signer] = recoverTransaction(signedTx);
           valid = signer.toLowerCase() === address.toLowerCase();
         } else {
--- a/dapps/react-dapp-v2/src/contexts/JsonRpcContext.tsx
+++ b/dapps/react-dapp-v2/src/contexts/JsonRpcContext.tsx
@@ -3,6 +3,7 @@ import { createContext, ReactNode, useContext, useState } from "react";
 import * as encoding from "@walletconnect/encoding";
 import { TypedDataField } from "@ethersproject/abstract-signer";
 import { Transaction as EthTransaction } from "@ethereumjs/tx";
+import { recoverTransaction } from "@celo/wallet-base";
 import {
   formatDirectSignDoc,
   stringifySignDocValues,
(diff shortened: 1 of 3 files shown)


dapps/react-dapp-v2/src/contexts/JsonRpcContext.tsx
```diff
@@ -252,9 +253,17 @@ export function JsonRpcContextProvider({
           },
         });
 
-        const valid = EthTransaction.fromSerializedTx(
-          signedTx as any
-        ).verifySignature();
+        let valid = false;
+        const [, reference] = chainId.split(":");
+        if (reference === "44787" || reference === "42220") {
```
bkrem, line 258: I realise the concrete chainIds aren't that simple to cross-reference the way they are specified here in the example dapp, but could we label these so they appear less like magic numbers? E.g.

```typescript
const CELO_ALFAJORES_CHAIN_ID = 44787

if (reference.toString() === CELO_ALFAJORES_CHAIN_ID) ...
```

dapps/react-dapp-v2/src/contexts/JsonRpcContext.tsx
```diff
@@ -252,9 +253,17 @@ export function JsonRpcContextProvider({
...
+          // Verify Celo transaction
+          const [, signer] = recoverTransaction(signedTx);
+          valid = signer.toLowerCase() === address.toLowerCase();
+        } else {
```
bkrem, line 262: When would we hit this `else` case if Celo testnet/mainnet are covered above? Does it even make sense to keep this logic if it doesn't actually work as expected as you indicated?

bkrem:
Thanks for putting in a fix here for Celo! A couple of initial questions, also do you have a specific recommended wallet you were testing this against?

dapps/react-dapp-v2/src/contexts/JsonRpcContext.tsx
kathaypacific: thanks for the suggestion, makes sense! i've updated and created celo variables :)

dapps/react-dapp-v2/src/contexts/JsonRpcContext.tsx
kathaypacific: maybe i misunderstand your question, but i think the `else` statement works for ethereum? i didn't want to touch the existing logic because i don't have the full context, i only noticed it wasn't working on Celo so wanted to fix that :) note that i am using a celo lib to recover the tx, which i'm not sure will work for other chains

kathaypacific:
> Thanks for putting in a fix here for Celo! A couple of initial questions, also do you have a specific recommended wallet you were testing this against?

hey @bkrem thanks for the review! i've made an update if you have time for another look. the wallet i used to test is Valora :)

dapps/react-dapp-v2/src/contexts/JsonRpcContext.tsx
bkrem: Oh yea woops my bad. Should've expanded the diff more, for some reason I thought this was a Celo specific function 🤦 

All good sorry for the confusion!

bkrem:
Thanks for the quick updates! 💯 

I get why it's being done but I'm not super keen on re-specifying specific chainIds inline in one of the functions, but I'll look into that on my side, i.e. how we can create a better way to reference by chainId based on the the constants/mappings we already have.

I will try to test and get this merged by EOD, otherwise first thing tomorrow EU time 👍

bkrem (approved):

PR #96, opened 2022-12-22
fix: dapps metadata

ganchoradkov:
PR to fix the issue where dapps were not providing `metadata` as part of the session proposal.
The problem was due to the dapps missing the appropriate  meta tags in `<head />`

While at it, also bumped the `@walletconnect` packages to latest & replaced the `QRModal` with `Web3Modal`


bkrem (approved):
Thanks for fixing + the modal upgrades 💯 

Tbh us doing this via manually calling `getAppMetadata()` [here](https://github.com/WalletConnect/web-examples/blob/main/dapps/react-dapp-v2/src/contexts/ClientContext.tsx#L258) and only then defaulting to the defaults was probably overkill but nice to have it properly configured via HTML metadata now 💪

ganchoradkov:
> Thanks for fixing + the modal upgrades 100
> 
> Tbh us doing this via manually calling `getAppMetadata()` [here](https://github.com/WalletConnect/web-examples/blob/main/dapps/react-dapp-v2/src/contexts/ClientContext.tsx#L258) and only then defaulting to the defaults was probably overkill but nice to have it properly configured via HTML metadata now muscle

@bkrem I digged into `getAppMetadata()` and even if it doesn't find anything still returns empty object so the fallback is kinda useless
```
    {
      name: "",
      description: "",
      url: "",
      icons: [""],
    } 
```

bkrem:
> even if it doesn't find anything still returns empty object

PR #102, opened 2023-01-11
feat: cosmos example dapp

ganchoradkov:
Updated the `cosmos` example dapp to use latest `@walletconnect`. 
Also added:
- universal-provider
- web3modal
- nextjs

Pending

- `universal-provider` release to replace canary


bkrem (approved):
A much needed revival of this example, thank you 💯

PR #103, opened 2023-01-13
refactor: web3wallet readme

ganchoradkov:
updated the web3wallet readme to fix the `metadata` usage


bkrem (approved):
Thanks for fixing 👍 no need to wait on me with obvious fixes like this btw 👌

PR #113, opened 2023-02-14
Add Emit Function

finessevanes:
- add emit example

### DIFF
--- a/dapps/react-dapp-v2/src/components/Header.tsx
+++ b/dapps/react-dapp-v2/src/components/Header.tsx
@@ -83,11 +83,11 @@ const Header = (props: HeaderProps) => {
             <Button outline color="black" onClick={ping}>
               {"Ping"}
             </Button>
+            <Button outline color="black" onClick={emit}>
+              {"Emit"}
+            </Button>
             <Button outline color="red" onClick={disconnect}>
               {"Disconnect"}
-            </Button>
-            <Button outline color="red" onClick={emit}>
-              {"Emit"}
             </Button>
           </SHeaderActions>
         </>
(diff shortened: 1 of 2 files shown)


dapps/react-dapp-v2/src/pages/index.tsx
```diff
@@ -108,6 +108,18 @@ const Home: NextPage = () => {
...
   };
 
+  async function emit() {
+    if (typeof client === "undefined") {
+      throw new Error("WalletConnect is not initialized");
+    }
+    
+    await client.emit({
+      topic: session?.topic || '',
+      event: { name: 'chainChanged', data: 'Hello World' },
```
bkrem, line 118: Think it's better to not provide `data` than using filler values for it

```suggestion
      event: { name: 'chainChanged' },
```

dapps/react-dapp-v2/src/components/Header.tsx
```diff
@@ -85,6 +86,9 @@ const Header = (props: HeaderProps) => {
             <Button outline color="red" onClick={disconnect}>
               {"Disconnect"}
             </Button>
+            <Button outline color="red" onClick={emit}>
```
bkrem, line 89: Would make this one black like ping and move it up one so we keep Disconnect on the furthest right (see the preview deployment)

bkrem (changes requested):
Thank you for adding a test emit button! 🙏 some minor notes, but already tested and it's logging in the example wallet as expected.

dapps/react-dapp-v2/src/pages/index.tsx
```diff
@@ -108,6 +108,18 @@ const Home: NextPage = () => {
...
   };
 
+  async function emit() {
+    if (typeof client === "undefined") {
+      throw new Error("WalletConnect is not initialized");
+    }
+    
+    await client.emit({
+      topic: session?.topic || '',
+      event: { name: 'chainChanged' },
```
bkrem, line 118: Looks like build failed bc `data` is `any` but still expected as a property.

```suggestion
      event: { name: 'chainChanged', data: {} },
```

bkrem (approved):

PR #116, opened 2023-02-15
chore(auth): update to auth-client@2.0.3, use rpc.walletconnect.com

bkrem:
## Context

- Incorporate fixes made in https://github.com/WalletConnect/auth-client-js/releases/tag/2.0.3
- Fix RPC queries done by the dapp example by using `rpc.walletconnect.com`


devceline (approved):
LGTM , but not sure if it's a regression from this PR but the UI text seems to overlap

image

bkrem:
> LGTM , but not sure if it's a regression from this PR but the UI text seems to overlap

Interesting, didn't touch UI here but can't actually repro that specific overlap even with the screen dimensions dev tooling. 

Think it's just general responsiveness issues that haven't been resolved on this sample.

PR #121, opened 2023-02-20
chore: update web3modal-standalone and signClient for react-dapp-v2

crypblizz8:
General package update for web3modal standalone + signClient + types + utils.

~~Also to see if this upgrade helps with a client wanting to see if their wallet appears.~~ Worked to resolve.


dapps/react-dapp-v2/src/contexts/ClientContext.tsx
```diff
@@ -154,7 +155,7 @@ export function ClientContextProvider({
           // Create a flat array of all requested chains across namespaces.
           const standaloneChains = Object.values(requiredNamespaces)
             .map((namespace) => namespace.chains)
-            .flat();
+            .flat() as string[];
```
crypblizz8, line 158: If you have other types suggestions, please let me know. 

I know we can use `SessionTypes.Struct.requiredNamespaces` but was unsure of how best to navigate it.

bkrem (approved):
Thanks for updating the dapp 💯 

Just tested with https://react-wallet.walletconnect.com/walletconnect, all good but the wallet also still needs to be updated. Could you add that here too pls? 🙏 If it requires a bunch of tweaks let's do it in a separate PR 👍

bkrem:
@crypblizz8 ping

PR #123, opened 2023-02-22
fix(dapp): removes reliance on `caip-api` pkg for blockchain data

bkrem:
- Context: https://walletconnect.slack.com/archives/C03RME3BS9L/p1677021657070509
- We've had some reports of some people seeing SSL errors when the dapp tries to resolve chain data via the `caip-api` pkg, which under the hood relies on the legacy `blockchain-api.xyz` as a data source.
- We still rely on a `blockchain-api.xyz` call to resolve some chain logos, but worst case scenario now is some broken logo assets if that domain can't be resolved, not a crashed dapp.

## How was this tested

- Explicitly blocked this domain in network inspector to recreate the issue and this fixes the crashing behaviour.

### DIFF
--- a/dapps/react-dapp-v2/package.json
+++ b/dapps/react-dapp-v2/package.json
@@ -24,7 +24,6 @@
     "axios": "^0.21.1",
     "blockies-ts": "^1.0.0",
     "bs58": "^5.0.0",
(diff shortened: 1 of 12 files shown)


ganchoradkov (approved):
:100: :100:

PR #127, opened 2023-03-05
Add Tezos Support

AndreasGassmann:
This PR adds Tezos support to 2 of the example projects.

There is also a PR to add the Tezos messages to the docs: https://github.com/WalletConnect/walletconnect-docs/pull/513


wallets/react-wallet-v2/src/views/SessionProposalModal.tsx
```diff
@@ -43,6 +45,7 @@ export default function SessionProposalModal() {
 
   // Add / remove address from EIP155 selection
   function onSelectAccount(chain: string, account: string) {
+    console.log('ACCOUNT SELECTED', chain, account)
```
bkrem, line 48: Leftover debug log

```suggestion

```

bkrem:
Apologies for the prettier noise, I've cleaned it up and synced this branch so the diff is clean now. Currently reviewing.

wallets/react-wallet-v2/src/utils/TezosRequestHandlerUtil.ts
```diff
@@ -0,0 +1,45 @@
...
+import { SignClientTypes } from '@walletconnect/types'
+import { getSdkError } from '@walletconnect/utils'
+
+export async function approveTezosRequest(
+  requestEvent: SignClientTypes.EventArguments['session_request']
+) {
+  const { params, id } = requestEvent
+  const { request } = params
+
+  const wallet = tezosWallets[request.params.account ?? Object.keys(tezosWallets)[0]] // TODO: Select correct wallet
```
bkrem, line 14: Remaining todo here, this currently always returns the values for the first account even if the second account is requested:

Screenshot 2023-03-09 at 14 43 38

dapps/react-dapp-v2/src/contexts/JsonRpcContext.tsx
```diff
@@ -1154,6 +1160,103 @@ export function JsonRpcContextProvider({
...
+                account: address,
+                payload,
+              },
+            },
+          });
+
+          return {
+            method: DEFAULT_TEZOS_METHODS.TEZOS_SIGN,
+            address,
+            valid: true,
```
bkrem, line 1250: Is there a way we can validate the signature here to evaluate `valid` properly?

wallets/react-wallet-v2/src/lib/TezosLib.ts
```diff
@@ -0,0 +1,109 @@
...
+    const prepared = await this.tezos.prepare.batch(
+      transaction.map((tx: any) => ({
+        amount: tx.amount,
+        to: tx.destination,
+        kind: tx.kind,
+        mutez: true
+      }))
+    )
+
+    const forged = await localForger.forge(prepared.opOb)
```
bkrem, line 97: Could you add context here please as to what is being done for this step and the semi-magical ones below (e.g. `new Uint8Array([3])` etc)

bkrem (changes requested):
Looking good overall, just some leftover tweaks AFAICT.

Thank you for adding tezos, was great meeting in Denver Andy! Sorry about the delay on review here, got pretty sick on the back end of ETHDenver and been recovering since.

wallets/react-wallet-v2/src/lib/TezosLib.ts
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  static async init({ mnemonic, path, curve }: IInitArguments) {
+    const params = {
+      mnemonic: mnemonic ?? Keyring.generateMnemonic(),
+      derivationPath: path ?? DEFAULT_PATH,
+      curve: curve ?? DEFAULT_CURVE
+    }
+
+    const Tezos = new TezosToolkit('https://mainnet.api.tez.ie')
```
AlirezaHaghshenas, line 58: So there is no way to `init` to a testnet? Is this intentional?

wallets/react-wallet-v2/src/utils/HelperUtil.ts
```diff
@@ -121,10 +121,17 @@ export function isElrondChain(chain: string) {
...
- export function isTronChain(chain: string) {
+export function isTronChain(chain: string) {
   return chain.includes('tron')
 }
 
+/**
+ * Check if chain is part of Tezos standard
+ */
+export function isTezosChain(chain: string) {
+  return chain.includes('tezos')
```
AlirezaHaghshenas, line 132: is it better to use `startsWith` instead of `includes`?

wallets/react-wallet-v2/src/utils/TezosRequestHandlerUtil.ts
AndreasGassmann: Currently, the request doesn't take any arguments, meaning that we do not know which account has to be shared. I thought that the wallet would just respond with all accounts that dApp has permissions for. Does that make sense, or would it be better if we added an the address to the params of the request?

wallets/react-wallet-v2/src/utils/HelperUtil.ts
AndreasGassmann: I copied the usage from the methods of the other chains, I guess if we change it here we should change it for all of them.

wallets/react-wallet-v2/src/lib/TezosLib.ts
AndreasGassmann: Tezos has certain prefixes for "forged" data. The "03" prefix means it's an operation. I will create a comment to point that out.

dapps/react-dapp-v2/src/contexts/JsonRpcContext.tsx
AndreasGassmann: It should be possible if we have access to the public key. But to have the public key, `tezos_getAccounts` has to be called before (or we have to fetch the public key from chain, which is not always possible). So it may not always be possible to do the validation. Should I still add it and it may not be possible sometimes, or just remove the "valid" flag?

wallets/react-wallet-v2/src/utils/TezosRequestHandlerUtil.ts
bkrem: > Currently, the request doesn't take any arguments

I'm a bit confused, for all the RPC methods other than `tezos_getAccounts` specifically the [spec](https://docs.walletconnect.com/2.0/advanced/rpc-reference/tezos-rpc#parameters-2) does have an `account` parameter and you do seem to already be passing the target account as a request parameter (e.g. [here](https://github.com/WalletConnect/web-examples/pull/127/files#diff-333319b492717cc438dc88afb2dadef1dbb71f3492dc04695a98830a8b4ac3a9R1204))

I see what you mean regarding `tezos_getAccounts`, since this kind of breaks out of the paradigm the wallet request handlers are built around (request is always directed at a specific account/signer).

In that case couldn't we simply do the following below to return all accounts?

```js
case TEZOS_SIGNING_METHODS.TEZOS_GET_ACCOUNTS:
    return formatJsonRpcResult(id, wallets.map(wallet => {
        algo: wallet.getCurve(),
        address: wallet.getAddress(),
        pubkey: wallet.getPublicKey()
    }))
```

dapps/react-dapp-v2/src/contexts/JsonRpcContext.tsx
bkrem: I see, would be great to have proper validation if you think it can be done here but not a showstopper for now 👍

wallets/react-wallet-v2/src/lib/TezosLib.ts
bkrem: +1, additionally I seem to get the following `500` when I try to perform `tezos_send`. Could of course just be because I don't have funds in the wallet Tezos mainnet but 500 seems like an odd return code for that:
Screenshot 2023-03-27 at 20 47 55

bkrem:
Sorry about the lag in response here @AndreasGassmann, I missed your initial responses to the review and only saw them end of last week.

Some small tweaks that I'd still recommend above, but beyond that if it works for you we can move ahead here and get this initial setup merged for the sake of getting this off a preview-only branch, and then carry on with deeper work where needed 👍

wallets/react-wallet-v2/src/lib/TezosLib.ts
AndreasGassmann: I quickly checked, I did not find how I can easily use the rpc for the correct network that is defined in `TezosData.ts`. It seems that other chains don't do this either.

dapps/react-dapp-v2/src/contexts/JsonRpcContext.tsx
AndreasGassmann: We can do this in the future (it's not as straight forward because of the reasons mentioned above), but as far as I can see all other protocols also have `valid: true` hardcoded as well.

AndreasGassmann:
I included one of your feedback, the others I left for now because they seem to be in line with how the other protocols are implemented, and I'm not sure how to easily add the feature.

In my opinion this can now be merged and we can improve it later on.

PR #129, opened 2023-03-07
chore(iss_chain-id): replaced hardcoded chainId with chainId from cacaoPayload

Cali93:
### Context

Updated the react-auth-wallet example to use the chainId from the cacao payload instead of the hardcoded one.
This will be merged once we bump the auth-client to 2.0.4 [(see  pending PR)](https://github.com/WalletConnect/auth-client-js/pull/63)


bkrem:
Thank you for already updating the examples 👍 let's attempt to redeploy once we have `2.0.4` published, then can approve once the deployments have been checked.

bkrem:
Oops I think I independently fixed the same thing without realising when I updated the auth examples to 2.0.4 😅

PR #137, opened 2023-03-17
fix:  chains in approved namespaces

ganchoradkov:
Fixed a reported issue where in the approved namespaces, `chains` prop was missing


bkrem (approved):
Oh wow so this was just an issue in the demo in the end? 🤦‍♂️ 

Thank you for looking into it and fixing 🙏

Issue #117, opened 2023-02-16
[dApp] Problem with smart contract wallet integration


basheerAudteye:
We have an NFT renting product, [rntd.io](https://rntd.io/) ,This product is still under development. We are still in the alpha version, we used walletConnect V1 to integrate our smart contract wallet with projects.
In V1, we did not face any problems with the integration with other projects, we used to authenticate personal_sign and typed_data to our smart contract with no effort due to verifySignature function that checks if the address is EOA address or smart contract address
https://github.com/WalletConnect/walletconnect-example-dapp/blob/master/src/helpers/utilities.ts#L181

But in V2, the provided dApp examples use  verifyEip155MessageSignature, which is only valid for EOA wallets, not for smart contract wallets that rely on the concept of erc1271.
https://github.com/WalletConnect/web-examples/blob/main/dapps/react-dapp-v2-with-ethers/src/pages/index.tsx#L69
When the function recovers the address of the wallet that signed the message, which owns the smart contract wallet, and compare it to the smart contract address, the validation will fail. As a result, we can not verify the personal message.
My question is, what are we missing, or what can we do to overcome this problem

basheerAudteye:
I made a pull request trying to fix the problem 
https://github.com/WalletConnect/web-examples/pull/140

PR #140, opened 2023-03-23
make personal sign supports non-custodial wallet

basheerAudteye:


dapps/react-dapp-v2/src/contexts/JsonRpcContext.tsx
```diff
@@ -1,4 +1,4 @@
-import { BigNumber, utils } from "ethers";
+import { BigNumber, getDefaultProvider, utils } from "ethers";
```
bkrem, line 1: Unused import:

```suggestion
import { BigNumber, utils } from "ethers";
```

dapps/react-dapp-v2/src/contexts/JsonRpcContext.tsx
```diff
@@ -292,10 +296,10 @@ export function JsonRpcContextProvider({
 
         // encode message (hex)
         const hexMsg = encoding.utf8ToHex(message, true);
-
+        const hashMsg = hashPersonalMessage(message)
```
bkrem, line 299: Since this is only used inside `verifySignature` at the bottom, can we call `hashPersonalMessage` inline there pls 🙏

bkrem (commented):
Closes #117 

Thank you for moving to fix this already @basheerAudteye 🙏 

Do you think it's worth also applying this to the `eth_sign` flow directly below?

dapps/react-dapp-v2/src/contexts/JsonRpcContext.tsx
basheerAudteye: > Closes #117
> 
> Thank you for moving to fix this already @basheerAudteye pray
> 
> Do you think it's worth also applying this to the `eth_sign` flow directly below?

in order to make non-custodial wallets fully functional with walletConnect V2 we need to apply the same logic to eth_signTypedData (eip-712) and eth_sign also

bkrem:
Superseded by https://github.com/WalletConnect/web-examples/pull/156 which includes your original commit here 👍 

Thanks again for raising and getting the ball rolling on the change 🙏

PR #142, opened 2023-03-24
Removed displaying version from index.tsx

mrakoczy-fp:
In ethers project displaying version caused an compilation error - package.json is not exported.


glitch-txs:
Thanks for the PR! This is not planned at the moment as we use some branches to test canary versions. Thanks!

PR #150, opened 2023-04-12
fix(relayer): restarts the relayer transport when region is changed

JonathanConn:
Created a seperate callback to handle region switching inside of react-wallet-v2 example


bkrem:
Could you merge from/rebase onto `main` for your branch (I just updated deps there)? 🙏 

I think there's definitely something funky up with `restartTransport` because at first it did nothing on the `react-wallet-v2` preview here, then after updating to latest `@walletconnect/*` deps I see the `Authorization error: JWT validation error: Invalid audience` issue when trying to change the region from default; cc @ganchoradkov

bkrem:
> After changing the region in the dropdown for first time, new socket is created correct but with the same URL

Yeah I saw the same thing initially, then ran against latest `@walletconnect/core` and saw the JWT error as mentioned above.

I just merged main into this branch so you'll see what I mean when the preview is redeployed.

JonathanConn:
> We have the same thing happening on the dapp side though, but thankfully the fix should be very similar (i.e. separate concerns between initialisation and restarting the transport): https://github.com/WalletConnect/web-examples/blob/main/dapps/react-dapp-v2/src/contexts/ClientContext.tsx#L282

I modified ClientContext to use restartTransport now as well. The implementation is slightly different because the client is being stored in a state. I temp downgraded the walletconnect deps to 2.4.10 to test without the jwt aud error. It worked correctly and I changed the deps back to 2.6.2

bkrem (approved):
I'm aware Gancho is still tinkering here, but approving since I think we're good to go on the concrete changes in the sample dapp/wallet

ganchoradkov (approved):
Awesome first PR uncovering hiccups in the JS sdk 💯

PR #153, opened 2023-04-25
Feat/examples upgrade

JonathanConn:
### Context
Adding chain switching to the react-wallet-v2 web-example

---
### Changes Description
All changes done in react-wallet-v2
- Modified UI to support chain switching by adding selectors and state in the store
- Updates state on change switch, and emits session updates using the sign client
- Additional helper method created to get full chain name from chainId

---
### Testing
Manually tested using the corresponding dapp

### DIFF
--- a/wallets/react-wallet-v2/src/utils/WalletConnectUtil.ts
+++ b/wallets/react-wallet-v2/src/utils/WalletConnectUtil.ts
@@ -43,7 +43,6 @@ export async function updateSignClientChainId(chainId: string, address: string) 
       }
     }
 
-    // console.log(session.namespaces ? { ...session.namespaces, ...newNamespace } : newNamespace)
 
     signClient.update({
       topic: session.topic,
--- a/wallets/react-wallet-v2/src/utils/WalletConnectUtil.ts
+++ b/wallets/react-wallet-v2/src/utils/WalletConnectUtil.ts
@@ -1,5 +1,4 @@
 import SignClient from '@walletconnect/sign-client'
-import { getFullChainName } from './HelperUtil'
 export let signClient: SignClient
 
 export async function createSignClient(relayerRegionURL: string) {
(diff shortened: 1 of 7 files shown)


wallets/react-wallet-v2/src/utils/WalletConnectUtil.ts
```diff
@@ -15,3 +15,39 @@ export async function createSignClient(relayerRegionURL: string) {
...
+          'eth_signTransaction',
+          'eth_sign',
+          'personal_sign',
+          'eth_signTypedData'
+        ],
+        events: ['chainChanged', 'accountsChanged']
+      }
+    }
+
+    // console.log(session.namespaces ? { ...session.namespaces, ...newNamespace } : newNamespace)
```
bkrem, line 46:
```suggestion
```

wallets/react-wallet-v2/src/utils/WalletConnectUtil.ts
```diff
@@ -15,3 +15,39 @@ export async function createSignClient(relayerRegionURL: string) {
...
+
+    signClient.update({
```
bkrem, line 48: I'm a bit confused about the intentions of this helper fn. `client.update()` is specifically to expand on the originally settled namespaces for the session, whereas "changing" chains/accounts is a superficial, UI-level event that won't alter the namespaces. I believe we need to use `client.emit()` for this, to simply signal this event to the peer client (i.e. the dapp in this case) so it can in turn update UI/own state. 

I think what we want to happen is very similar to what the sample wallet client is doing here in the UniversalProvider tests (see the helper above for the `accountsChanged` equivalent), where `client.emit(...)` is called when the chainId is updated:
https://github.com/WalletConnect/walletconnect-monorepo/blob/v2.0/providers/universal-provider/test/shared/WalletClient.ts#L111

cc @ganchoradkov

bkrem (changes requested):
Functionality-wise looking great (did some testing on my end), but think we still need to address what exactly the client is doing on `chainChanged` + avoid the currently "lossy" forwarding of the chainId 👍

bkrem (approved):
Nice 💪

I like the approach of explicitly doing `update` if the targeted chain isn't already covered in namespaces. Simply letting the emit fail on validation in this case would also be a valid option since we shouldn't technically be able to switch to chain that isn't agreed for the session.

PR #156, opened 2023-04-28
fix(dapp): makes sig validation compatible with ERC-1271

### LINKED ISSUE #117: [dApp] Problem with smart contract wallet integration

bkrem:
- Fixes #117 
- Supersedes #140 
- Tested that existing signature validation still works as expected with https://react-wallet.walletconnect.com, on top of the additional ERC/EIP 1271 validation now accounted for.

### DIFF
--- a/dapps/react-dapp-v2/src/contexts/JsonRpcContext.tsx
+++ b/dapps/react-dapp-v2/src/contexts/JsonRpcContext.tsx
@@ -1,7 +1,6 @@
-import { BigNumber, getDefaultProvider, utils } from "ethers";
+import { BigNumber, utils } from "ethers";
 import { createContext, ReactNode, useContext, useState } from "react";
 import * as encoding from "@walletconnect/encoding";
-import { TypedDataField } from "@ethersproject/abstract-signer";
 import { Transaction as EthTransaction } from "@ethereumjs/tx";
 import { recoverTransaction } from "@celo/wallet-base";
 import {
@@ -26,6 +25,7 @@ import {
   formatTestTransaction,
   getLocalStorageTestnetFlag,
   hashPersonalMessage,
+  hashTypedDataMessage,
   verifySignature,
 } from "../helpers";
 import { useWalletConnectClient } from "./ClientContext";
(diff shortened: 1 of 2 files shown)


dapps/react-dapp-v2/src/contexts/JsonRpcContext.tsx
```diff
@@ -315,10 +317,14 @@ export function JsonRpcContextProvider({
           throw new Error(`Missing chain data for chainId: ${chainId}`);
         }
 
-        const valid = _verifyEip155MessageSignature(
-          message,
+        const hashMsg = hashPersonalMessage(message);
+        const rpc =
```
Cali93, line 321: Should we throw an error if there's no rpcProvider for the given chainId ?

Cali93 (approved):
LGTM, just a minor question around RPC url.

dapps/react-dapp-v2/src/contexts/JsonRpcContext.tsx
```diff
@@ -315,10 +317,14 @@ export function JsonRpcContextProvider({
...
+          rpcProvidersByChainId[Number(chainId.toString().split(":")[1])];
```
Cali93, line 322: Should we throw an error if there's no rpcProvider for the given chainId ?

Cali93 (approved):
LGTM, just a minor question around RPC

dapps/react-dapp-v2/src/contexts/JsonRpcContext.tsx
bkrem: Good call thank you. Actually helped me clear up the existing check there which was redundant 👍

PR #162, opened 2023-05-05
Add XRPL example

tequdev:
## Add following examples
- react-dapp-v2
- react-web3wallet

## XRPL Reference
https://docs.walletconnect.com/2.0/advanced/rpc-reference/xrpl-rpc


bkrem:
Hi @develoQ,

Did some initial testing between these previews of the sample dapp and wallet.

Looks like I'm getting "Account not found" when I try to approve e.g. `xrpl_signTransaction`:
Screenshot 2023-05-12 at 14 48 52

Happens with both accounts if I establish a session with two active accounts.

bkrem (changes requested):
See issue above

develoQ:
Thanks for the review.

XRP Ledger requires a reserve in order to use your account.
You can get it from Facuet on Testnet.

Regarding your point of view, we can address the following.

1. facuet the account created by the user himself
    https://test.bithomp.com/faucet/

2. automatically deposit from facuet to the account created in the seed creation process (additional code is needed)

dapps/react-dapp-v2/src/contexts/JsonRpcContext.tsx
```diff
@@ -1265,6 +1270,123 @@ export function JsonRpcContextProvider({
...
+                  TransactionType: "OfferCreate",
+                  Account: address,
+                  Flags: 524288,
+                  TakerGets: "15000000000",
+                  TakerPays: {
+                    currency: "USD",
+                    issuer: "rvYAfWj5gh67oV6fW32ZzP3Aw4Eubs59B",
+                    value: "7072.8",
+                  },
+                },
```
antondalgren, line 1300: Could we consider using a simpler TX e.g. a regular Payment. 
```javascript
              params: {
                tx_json: {
                  Account: address,
                  Destination: "rTooLkitCksh5mQa67eaa2JaWHDBnHkpy",
                  Amount: "1",
                  TransactionType: "Payment",
                },
                submit: true,
                autofill: true,
              },
```

Also keep the amounts to a minimum value to not drain a wallet.

antondalgren:
Would it also be possible to set the methods to the optional namespace? If the wallet has support for one, or both methods, it should still be ok to connect.

dapps/react-dapp-v2/src/contexts/JsonRpcContext.tsx
develoQ: Fixed for `xrpl_signTransaction`.
How do you feel about `xrpl_signTransaction_for`?

TiltUp-Rad:
the fact that somebody is looking into is amazing. info about XRP regarding web3modal and wallet connect is super scarce!
cheers!

glitch-txs:
Hi @tequdev should we revisit this PR? can we fix the conflicts as well? Thanks

tequdev:
> Hi @tequdev should we revisit this PR? can we fix the conflicts as well? Thanks

I'd like to try to resolve conflicts.
Where has react-web3wallet migrated to?

glitch-txs:
https://github.com/WalletConnect/web-examples/tree/main/wallets

tequdev:
replaced to #340

PR #172, opened 2023-06-02
make react-dapp-v2-with-ethers support non-custodial wallet

basheerAudteye:


basheerAudteye:
@chadyj @bkrem 
can you review the following pull request? It is similar to this [commit](https://github.com/WalletConnect/web-examples/commit/c6e9c2d6149f05617f0125dccc3ac04a062540c8)  but applied to a different model

PR #202, opened 2023-06-26
chore: update toe latest 1.8.2

crypblizz8:
Latest update for w3w SDK.


bkrem (approved):
Thank you 🙏 

This was effectively already updated earlier today via renovateBot here: https://github.com/WalletConnect/web-examples/pull/197

But because of the `^` in the version it only updated the lockfile apparently (`^` = any minor/patch version higher than what comes after the `^` is acceptable).

Should also update `package.json` itself now that you removed the `^` caret here 👍

PR #204, opened 2023-06-27
Test: buildApprovedNamespaces in our wallet example

crypblizz8:
Test nameSpaceBuilder in our wallet example

Note here: i did not update the UI. So optional does not show up

image

Refer to the console log: `buildApprovedNameSpace data:` that is called after you press approve


bkrem:
@crypblizz8 can we close this or is there further action to be taken?

PR #260, opened 2023-08-01
Feat/e2e Add More Tags

JonathanConn:
Final dom tags needed for e2e testing + updated docs


bkrem (approved):
Removed the package-lock.json that was added here since we already have yarn.lock 👍

PR #294, opened 2023-09-19
refactor: updated `sign` to `web3wallet`

ganchoradkov:
- updated main wallet example to use `web3wallet` instead of `sign-client`
- implemented Auth request handling
- removed `example-wallet-eip155` as it's redundant
- removed the old `web3wallet` example as it was outdated
- fixed bug preventing approval of session proposals with only optional namespaces

TODO: after merging to remove `react-web3wallet` vercel deployment


wallets/react-wallet-v2/yarn.lock
```diff
@@ -1903,7 +1903,7 @@
     "@stablelib/constant-time" "^1.0.1"
     "@stablelib/wipe" "^1.0.1"
 
-"@stablelib/sha256@1.0.1":
+"@stablelib/sha256@1.0.1", "@stablelib/sha256@^1.0.1":
```
0xAsimetriq, line 1906: Looks like 2 similarish versions of same lib here, maybe lockfile needs to be deleted and re-generated?

0xAsimetriq (approved):

wallets/react-wallet-v2/src/pages/session.tsx
```diff
@@ -18,15 +18,15 @@ export default function SessionPage() {
...
+  const { activeChainId } = useSnapshot(SettingsStore.state)
 
   useEffect(() => {
     if (query?.topic) {
       setTopic(query.topic as string)
     }
   }, [query])
 
-  const session = signClient.session.values.find(s => s.topic === topic)
+  const session = web3wallet.engine.signClient.session.values.find(s => s.topic === topic)
```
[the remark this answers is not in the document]
ganchoradkov: We don't `sessions` api exposed directly and will add it 💯

wallets/react-wallet-v2/yarn.lock
ganchoradkov: I tried it but it's regenerated the same way. May be inner dependencies have it declared differently

PR #295, opened 2023-09-21
feat: UI update

ganchoradkov:
UI update on the request modals in `react-wallet-v2`

- all request modals updated with new design
- added blocker screen when connecting with `isScam` flagged dapp
- implemented `buildApprovedNamespaces` util instead of manual namespace parsing 

- dependent on https://github.com/WalletConnect/web-examples/pull/294


boidushya (approved):
LGTM!

PR #300, opened 2023-09-29
Check for the whole chain identifier ( namespace + ":" + reference )

radumojic:
Check for the whole chain identifier ( `namespace` + ":" + `reference` ) in `getChainData` instead of just reference(chainId) to avoid conflicts between chains that share the same reference ( `testnet` for NEAR Testnet and Tezos Testnet or `1` for Ethereum Mainnet and MultiversX Mainnet)

Updated the MultiversX namespace to the CAIP-2 compliant `mvx` used in the ecosystem


bkrem (approved):
Thank you! 🙏

PR #321, opened 2023-11-15
chore: updates deps to 2.10.5

ganchoradkov:


bkrem:
@ganchoradkov we still want this? Not actually clear to me why renovatebot hasn’t opened PRs for these updates already

glitch-txs (approved):

PR #332, opened 2023-12-01
chore: update `@celo/wallet-base` dependency

bakoushin:
Update outdated `@celo/wallet-base` dependency (3.0.1 --> 5.1.1).

In the WalletConnect demo dApp, it's used for verifying the correctness of the `eth_signTransaction` result using Celo network wallet (e.g. [Valora wallet](https://valoraapp.com)).

The latest version of `@celo/wallet-base` supports transactions encoded in the [CIP-42](https://github.com/celo-org/celo-proposals/blob/master/CIPs/cip-0042.md) format, which is becoming the recommended one in within the Celo network. The legacy Celo transaction types are still supported as well.

### DIFF
--- a/dapps/react-dapp-v2/package.json
+++ b/dapps/react-dapp-v2/package.json
@@ -10,7 +10,7 @@
     "prettier:write": "prettier --write '**/*.{js,ts,jsx,tsx}'"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


bkrem (approved):
Just tested with Valora against the dapp preview, LGTM ✅ 

Thanks for updating this 🙏

PR #341, opened 2023-12-11
experimental: Goerli Smart Accounts

tomiir:
Credit to @svenvoskamp @0xAsimetriq 

## Summary
- Adds `SmartAccount.lib` and `SmartAccountCard`
- If on Goerli testnet, generates a Safe Smart account for a given private key using Pimlico's bundler.
- Displays Smart Account id.

## Notes
- Rebased with master, but can do a smaller change pr to @0xAsimetriq if that's better for CR.

## WIP
- Abstracting aux functions instead of having all inside of `getSmartAccount`
- Adding `useSmartAccount` hook
- Allow configuring the network (Sepolia and Goerli for starters)


0xAsimetriq (changes requested):
Great start on this 💪

There is currently one big change that needs to be address, which is deletion of all those additional non-evm chains and associated packages. This was done during hackweek in order to simplify the example for our partners, but all of them should still remail and function for the actual `main` branch.

tomiir:
Closing in favor of: #342

PR #342, opened 2023-12-13
experimental: Goerli Smart Account

tomiir:
Replaces #341 

Credit to @svenvoskamp @0xAsimetriq 

## Summary
- Adds `SmartAccount.lib` and `SmartAccountCard`
- If on Goerli testnet, generates a Safe Smart account for a given private key using Pimlico's bundler.
- Displays Smart Account id.

## Notes
- Pimlico released a new version of `permisionless` but it breaks on Goerli

### DIFF
--- a/wallets/react-wallet-v2/package.json
+++ b/wallets/react-wallet-v2/package.json
@@ -39,6 +39,7 @@
     "mnemonic-keyring": "1.4.0",
     "near-api-js": "^0.45.0",
     "next": "12.1.5",
+    "permissionless": "0.0.0-main.20231120T173047",
     "react": "17.0.2",
     "react-code-blocks": "0.0.9-0",
     "react-dom": "17.0.2",
@@ -46,15 +47,16 @@
     "react-qr-reader-es6": "2.2.1-2",
     "solana-wallet": "^1.0.2",
     "tronweb": "^4.4.0",
-    "valtio": "1.6.0"
+    "valtio": "1.6.0",
+    "viem": "^1.19.13"
   },
   "devDependencies": {
     "@types/node": "17.0.35",
-    "@types/react": "18.0.9",
+    "@types/react": "17.0.2",
     "eslint": "8.15.0",
     "eslint-config-next": "12.1.6",
     "eslint-config-prettier": "8.5.0",
     "prettier": "2.6.2",
-    "typescript": "4.6.4"
+    "typescript": "5.2.2"
   }
 }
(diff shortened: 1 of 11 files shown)


arein (approved):
LGTM!
If I don't have enough funds the experience just gets stuck. I think some error handling is missing. But that can be done in a follow up PR

wallets/react-wallet-v2/package.json
```diff
@@ -39,22 +39,24 @@
     "mnemonic-keyring": "1.4.0",
     "near-api-js": "^0.45.0",
     "next": "12.1.5",
+    "permissionless": "0.0.0-main.20231120T173047",
```
0xAsimetriq, line 42: I believe that everything included in this version should be available on latest one of permisionless by now, you might be able to use proper version here in that case.

wallets/react-wallet-v2/package.json
```diff
@@ -39,22 +39,24 @@
...
     "react": "17.0.2",
     "react-code-blocks": "0.0.9-0",
     "react-dom": "17.0.2",
     "react-hot-toast": "^2.4.1",
     "react-qr-reader-es6": "2.2.1-2",
     "solana-wallet": "^1.0.2",
     "tronweb": "^4.4.0",
-    "valtio": "1.6.0"
+    "valtio": "1.6.0",
+    "viem": "^1.19.13"
```
0xAsimetriq, line 51: nit: I'm not sure about this repo exactly, but we tend to avoid using `^` and other "relaxed" modifiers in npm versions and instead prefer to use strict versioning. Might be good idea to apply here (needs to remove package lock file and re-install dependencies fresh)

wallets/react-wallet-v2/src/components/SmartAccountCard.tsx
```diff
@@ -0,0 +1,137 @@
...
+  name,
+  logo,
+  rgb,
+  address = '',
+  chainId,
+  isActiveChain
+}: Props) {
+  const [copied, setCopied] = useState(false)
+  const { activeChainId } = useSnapshot(SettingsStore.state)
+  const { deploy, isDeployed, address: smartAccountAddress, loading, sendTestTransaction } = useSmartAccount(eip155Wallets[address].getPrivateKey() as `0x${string}`)
```
0xAsimetriq, line 31: Maybe break this down into multiline to avoid long single line.

0xAsimetriq (approved):
Not much from me to add here besides few small comments, solid job 💪

wallets/react-wallet-v2/package.json
tomiir: I tried updating yesterday and got signature mismatch errors with the same code that works 😵 
Will try again today, maybe I need to tweak something for it to work.

tomiir:
Sorry about this but need to do a new PR because of all this reorg 🙏 
#347

PR #351, opened 2023-12-15
feat: Smart Account relay connection + paymaster base

tomiir:
## Summary
- Adds necessary logic to enable the wallet's Goerli SmartAccount usage via the WC relay 🚀 
- Adds Verifying Paymaster implementation, with a settings field where it can be toggled on/off if user decides to pay with native token. 🔍 
- Adds base code for ERC20 Paymaster integration.

## Notes
- Can use https://react-app.walletconnect.com/ to test it out! 
- `eth_signTransaction` does not currently work as Pimlico's smart client does not expose the funcitonality and viem's base client forbids smart accounts from signing transactions.


arein:
Screenshot 2024-01-31 at 11 51 16 PM

Awesome work! I think I found a small bug but don't think that needs to block this PR.

Repro steps:
1. deploy smart account on sepolia
2. connect https://lab.web3modal.com/library/ethers/
3. see console error above

arein (approved):

PR #402, opened 2024-01-10
feat: loading indicators

ganchoradkov:
- Added loading indicators in main wallet example on all awaits
- Added loading indicators/modals in main dapp example on disconnect & proposing session on known pairing
- General house keeping - removing unused code/unused imports

### DIFF
--- a/advanced/wallets/react-wallet-v2/src/views/SessionProposalModal.tsx
+++ b/advanced/wallets/react-wallet-v2/src/views/SessionProposalModal.tsx
@@ -1,12 +1,11 @@
-import { Col, Divider, Grid, Row, Text, styled } from '@nextui-org/react'
-import { Fragment, useCallback, useMemo } from 'react'
+/* eslint-disable react-hooks/rules-of-hooks */
+import { Col, Grid, Row, Text, styled } from '@nextui-org/react'
+import { useCallback, useMemo, useState } from 'react'
 import { buildApprovedNamespaces, getSdkError } from '@walletconnect/utils'
 
 import DoneIcon from '@mui/icons-material/Done'
 import CloseIcon from '@mui/icons-material/Close'
 
-import ProjectInfoCard from '@/components/ProjectInfoCard'
-import RequestModalContainer from '@/components/RequestModalContainer'
 import ModalStore from '@/store/ModalStore'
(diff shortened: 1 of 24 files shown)


glitch-txs (approved):

advanced/wallets/react-wallet-v2/src/views/SessionProposalModal.tsx
```diff
@@ -246,39 +229,54 @@ export default function SessionProposalModal() {
...
         return
       }
     }
+    setIsLoadingReject(false)
     ModalStore.close()
-  }
+  }, [proposal])
 
-  return (
+  return !proposal ? (
```
bkrem, line 266: One thing I noticed here is that reusing a spent pairing URI will keep this in infinite load but I remember us discussing yesterday this really being more of a protocol-level problem with awaiting the peer response right?

advanced/wallets/react-wallet-v2/src/views/SessionSendTransactionModal.tsx
```diff
@@ -1,21 +1,19 @@
-import { Fragment, useState } from 'react'
-import { Divider, Modal, Text } from '@nextui-org/react'
+/* eslint-disable react-hooks/rules-of-hooks */
```
bkrem, line 1: Why?

bkrem (approved):
Nice much clearer in the UI with these tweaks 🙌

advanced/wallets/react-wallet-v2/src/views/SessionProposalModal.tsx
ganchoradkov: it would actually keep the loading for ~5 mins until the pairing expires and show an error message. It's not ideal but for used pairings, we can't do much. Once improvement that we could do, is track used pairings on the instance and throw immediately if we already used it. If pairing was used on different instance/wallet, we're stuck with current approach

advanced/wallets/react-wallet-v2/src/views/SessionSendTransactionModal.tsx
ganchoradkov: fixed it. The initial modal setup required rework of the modal parameters as it was causing conditional hooks usage

arein (approved):

PR #405, opened 2024-01-12
feat: wallet connection via deeplink

ganchoradkov:
Added new page `/wc` that accepts deeplinks (`?uri=wc%3Ab2222ed98bec37db97d17028849eb3a3d0d5179cac4cfc036e6602d3e32bd87f%402%3Frelay-protocol%3Dirn%26symKey%3D3081499b0c00c3674b648f3d3262ae8aa689639166a0c5ba69f1cad24d171252`) to initialize connection instead of only copy/paste of qr


arein (approved):
Does this handle `/wc`?

PR #211, opened 2023-06-30
feat: adds zk sync era

githubdoramon:
Adds zkSync Era to dapp-v2, wallet-v2 and web3wallet

PR #415, opened 2024-01-16
feat: add Astar zKatana testnet

codespool:
- Add Astar zKatana chain data to
  - `react-dapp-v2`
  - `react-dapp-v2-with-ethers`
  - `react-dapp-v2-with-web3js`

Modify `EIP155Colors`, `EIP155ChainData`, `EIP155Metadata` objects, and add logo image to `./public/assets`


bkrem (changes requested):
Nice, thank you for adding zKatana to the sample dapps 🙏 

All looks good to me on that side of the equation, but we're currently missing support on the sample wallet side (see screenshot).

Since this is an EIP155/EVM chain, adding it to the wallet should be just as simple. Here's an example from when the zkSync team added support to the wallet: https://github.com/WalletConnect/web-examples/pull/211

Screenshot 2024-01-17 at 14 31 19

bkrem:
Adding @glitch-txs and @ganchoradkov as co-reviewers for visibility

codespool:
Hi @bkrem , sorry for the delay.
Added the zKatana config to wallet packages.

PR #440, opened 2024-02-01
fix: reintroduces tezos & cosmos implementations

ganchoradkov:
while resolving the install/build issues caused by the outdated `mnemonic-keyring` dep


bkrem (approved):
Thanks for fixing 💯

PR #441, opened 2024-02-01
Display smart account address on permission modal with eoa wallet add…

Bacis:
Screenshot 2024-02-01 at 14 10 08

On requested permission modal, in case there's a smart account deployed, display its address right under the wallet address.

### DIFF
--- a/advanced/wallets/react-wallet-v2/src/views/SessionProposalModal.tsx
+++ b/advanced/wallets/react-wallet-v2/src/views/SessionProposalModal.tsx
@@ -8,7 +8,7 @@ import CloseIcon from '@mui/icons-material/Close'
 
 import ModalStore from '@/store/ModalStore'
 import { cosmosAddresses } from '@/utils/CosmosWalletUtil'
-import { eip155Addresses } from '@/utils/EIP155WalletUtil'
+import { eip155Addresses, eip155Wallets } from '@/utils/EIP155WalletUtil'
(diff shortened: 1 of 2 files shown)


tomiir:
Looks good, but I'd check https://github.com/WalletConnect/web-examples/pull/441 first since it has some changes that would make this a bit simpler.
 I had already worked a bit on this screen so it might get conflicts, let me know if you need any help 🙏 
@Bacis

advanced/wallets/react-wallet-v2/src/views/SessionProposalModal.tsx
```diff
@@ -216,6 +218,29 @@ export default function SessionProposalModal() {
...
+            break;        
+          case 11155111:
+            enabledChain = 'sepolia'
+            break;
+        }
+
+        const pKey = eip155Wallets[getAddress(chain?.namespace) as `0x${string}`].getPrivateKey() as `0x${string}`
+        const smartAccountClient = new SmartAccountLib(pKey, enabledChain as SmartAccountEnabledChains)
+      
+        smartAccountClient.getSmartAccountAddress().then(res => {
```
ganchoradkov, line 237: is there a chance that `getSmartAccountAddress` fails for some reason? 
Simple catch and log of the error might prevent it causing issues with other functionality in the wallet

ganchoradkov (approved):
💯

PR #460, opened 2024-02-05
fix: session approval for non evm chains

ganchoradkov:
fixing a bug where proposal containing only non-evm chains results in exception


bkrem (approved):
💯

PR #469, opened 2024-02-09
hotfix for wallet3inbox connection

Bacis:
https://files.slack.com/files-pri/TB3H1FRHU-F06JGM2L94H/screenshot_2024-02-09_at_12.17.33___pm.png

Connection on wallet3inbox hanged on approval. 

Added error handling and validation to skip SA connection when there are none

### DIFF
--- a/advanced/wallets/react-wallet-v2/src/lib/SmartAccountLib.ts
+++ b/advanced/wallets/react-wallet-v2/src/lib/SmartAccountLib.ts
@@ -39,8 +39,6 @@ export class SmartAccountLib {
       throw new Error('A Pimlico API Key is required')
     }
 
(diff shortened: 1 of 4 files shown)


advanced/wallets/react-wallet-v2/src/lib/SmartAccountLib.ts
```diff
@@ -39,8 +39,6 @@ export class SmartAccountLib {
       throw new Error('A Pimlico API Key is required')
     }
 
-    console.log('investigate 2', chain, privateKey, sponsored)
```
arein, line 42: love me some private Key logging :)

arein (approved):
Icing on top would be a unit/functional tests so we prevent regessions but as a hotfix this is great

PR #479, opened 2024-02-12
connect to a dApp with SA, create SA on building a crypto link

Bacis:
Screenshot 2024-02-12 at 12 05 06

Filtered only allowed chains when connecting SA. If SA is available, make sure it is the one that gets displayed on the dApp you connect.


arein (approved):
extra points for a Unit Test to cover regressions

also consider adding a toggle for this in settings such that we can integration test both cases

PR #481, opened 2024-02-13
fix: smart account feature flag

lukaisailovic:
Change the order of accounts inside namespace, making smart account last instead of first


tomiir (approved):
Looks good!

PR #482, opened 2024-02-14
Feat/sa paymaster usdc

Bacis:


advanced/wallets/react-wallet-v2/src/hooks/useSmartAccount.ts
```diff
@@ -26,6 +26,8 @@ export default function useSmartAccount(signerPrivateKey: Hex, chain: Chain) {
         if (e instanceof TransactionExecutionError) {
           // shorten the error message
           styledToast(e.cause.message, 'error')
+        }  else if (e instanceof RpcRequestError) {
+          styledToast("User should have at least 34 USDC in Smart Account wallet to execute the first transaction.", 'error')
```
tomiir, line 30: Is 34 the actual limit? (such a weird number 🤔 )
Also, is this the only cause for an RPCRequestError?

Bacis: I think its better to stick to errors that come from pimlico. There might be changes on their side and maybe users won't need to have such insider logic.

PR #483, opened 2024-02-14
vue sample

mabsattar:
Added a vue sample dapp for the web-examples.


dapps/web3modal/vue/src/components/HelloWorld.vue
```diff
@@ -0,0 +1,58 @@
+<template>
```
glitch-txs, line 1: I think we can remove this component

mabsattar: Sure, I've removed it

glitch-txs (approved):

glitch-txs:
thanks!

PR #484, opened 2024-02-17
add wagmi/view 2.x

0xPr0f:
add wagmi/view 2.x


glitch-txs (approved):
Where you able to run it? Will be merging it into `feat/react-tx` so I can take a closer look, then to main.

Thanks! :)

PR #492, opened 2024-02-22
feat: pending requests counter

ganchoradkov:
Added counter for pending requests with timer of the remaining expiry in session details
image


arein (approved):

advanced/wallets/react-wallet-v2/src/pages/session.tsx
```diff
@@ -35,6 +36,20 @@ export default function SessionPage() {
...
+    setPendingRequests(requestsForSession)
+  }, [session])
+
+  useEffect(() => {
+    const interval = setInterval(() => {
+      getPendingRequests()
+    }, 1000)
+    getPendingRequests()
+    return () => clearInterval(interval)
+  }, [getPendingRequests])
```
arein, line 52: why don't we have events for this?

ganchoradkov: we do, but this bundles a couple of functionalities in single command (the interval)
- refreshes the list when when new request comes in
- refreshes the list when request is responded to
- refreshes the expiry timer
any other approach would be more lengthy/complex to develop

PR #495, opened 2024-03-01
fix: unknown validation

ganchoradkov:
Fixed flaky verify `unknown` validation test case in react dapp by explicitly removing the iframe thus preventing attestations from being POSTed. The old behaviour was working fine until SDKs stopped accepting custom verify URLs.

The loop is needed as Verify tries to init new iframe(with different urls) multiple times


bkrem:
Is this related to the recent reports re:CoWSwap etc or something you found independently while debugging that?

advanced/dapps/react-dapp-v2/src/contexts/ClientContext.tsx
```diff
@@ -320,6 +317,22 @@ export function ClientContextProvider({
     origin,
   ]);
 
+  useEffect(() => {
+    const claimedOrigin =
+      localStorage.getItem("wallet_connect_dapp_origin") || origin;
+    let interval: NodeJS.Timer;
+    // simulates `UNKNOWN` validation by removing the verify iframe thus pereventing POST message
```
bkrem, line 324:
```suggestion
    // simulates `UNKNOWN` validation by removing the verify iframe thus preventing POST message
```

bkrem: Could we elaborate on why an interval is needed here for context. You mentioned in the PR description but not intuitive here, e.g. because we try to create an iframe on .com and .org

bkrem (approved):

lukaisailovic:
Lets make sure the tests are passing before merge

advanced/dapps/react-dapp-v2/src/contexts/ClientContext.tsx
```diff
@@ -320,6 +317,22 @@ export function ClientContextProvider({
     origin,
   ]);
 
+  useEffect(() => {
+    const claimedOrigin =
+      localStorage.getItem("wallet_connect_dapp_origin") || origin;
```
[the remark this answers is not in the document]
ganchoradkov: Yes, so the page & client can be refreshed with intended data/behaviour

PR #511, opened 2024-03-14
feat: safe smart accounts

lukaisailovic:
Create single SmartAccountLib instance that should be extended for each implementation. Added modular code to support multiple implementations


advanced/wallets/react-wallet-v2/src/lib/smart-accounts/SmartAccountLib.ts
```diff
@@ -0,0 +1,174 @@
...
+    })
+
+    this.bundlerClient = createClient({
+      transport: this.bundlerUrl,
+      chain: this.chain
+    })
+      .extend(bundlerActions)
+      .extend(pimlicoBundlerActions)
+
+    const name = this.constructor.name.replace('SmartAccountLib', '')
```
tomiir, line 89: nice trick 😎

advanced/wallets/react-wallet-v2/src/lib/smart-accounts/SmartAccountLib.ts
```diff
@@ -0,0 +1,174 @@
...
+      chain: this.chain.name,
+      type: this.type
+    })
+    this.initialized = true
+  }
+
+  getMnemonic(): string {
+    throw new Error('Method not implemented.')
+  }
+  getPrivateKey(): string {
```
tomiir, line 109: what happens if this is instanced with a signer?

lukaisailovic: What do you mean by signer?

tomiir (approved):
Looks good. Might be a good moment to update to permisionless 0.1.0 too

advanced/wallets/react-wallet-v2/src/lib/smart-accounts/SmartAccountLib.ts
tomiir: instead of using a privatekey you can instantiate the smart account client using a signer (e.g. the magic signer)

lukaisailovic: Currently we only instantiate this with a private key. Impossible to do so in a different manner. If there is a need, we can extend this to support mnemonics as well. Left that method because I introduced `EIP155Wallet` interface in the last PR as well

PR #512, opened 2024-03-18
fix: catch namespace builder exception

ganchoradkov:
`buildApprovedNamespaces` throws when required namespaces are not supported by the wallet and must be catched to avoid unhandled exceptions

### DIFF
--- a/advanced/wallets/react-wallet-v2/src/views/SessionProposalModal.tsx
+++ b/advanced/wallets/react-wallet-v2/src/views/SessionProposalModal.tsx
@@ -226,17 +226,22 @@ export default function SessionProposalModal() {
     }
   }, [])
 
-  const namespaces = buildApprovedNamespaces({
-    proposal: proposal.params,
-    supportedNamespaces
-  })
+  const namespaces = useMemo(() => {
+    try {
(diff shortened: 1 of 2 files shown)


advanced/wallets/react-wallet-v2/src/views/SessionProposalModal.tsx
```diff
@@ -226,17 +226,22 @@ export default function SessionProposalModal() {
...
-    supportedNamespaces
-  })
+  const namespaces = useMemo(() => {
+    try {
+      // the builder throws an exception if required namespaces are not supported
+      return buildApprovedNamespaces({
+        proposal: proposal.params,
+        supportedNamespaces
+      })
+    } catch (e) {}
```
arein, line 236: does this need a toast?

arein (approved):

advanced/wallets/react-wallet-v2/src/views/SessionProposalModal.tsx
ganchoradkov: if namespaces are not built there is message in the proposal
image

PR #515, opened 2024-03-19
fix: smart account display

lukaisailovic:
Display smart accounts from the list of available accounts.
Fix the settings key for safe


tomiir (approved):
LGTM!

PR #533, opened 2024-04-08
fix: no supported chains

ganchoradkov:
- fixed unhandled exception when `allowedChainIds` is empty due all chains in the proposal are not supported by the wallet
- now we're disabling the approve button when the wallet doesn't support any of the proposed chains


arein (approved):
Thank you!

PR #538, opened 2024-04-16
fix: session approval accounts order

ganchoradkov:
fixed a bug where when smart accounts were enabled, they were replacing the available EOA on proposal approval


arein (approved):
Would be nice to have a unit or functional test so this won't break again

PR #540, opened 2024-04-18
 added CAIP-25 to remove the round-trip required for wallet_getCapabilities

KannuSingh:
on session approval, wallet capabilities data as defined for wallet_getCapabilities in EIP5792 is added to sessionProperties,
this gives another way for dapps to check wallet's capabilities without requesting via wallet_getCapabilities method.

### DIFF
--- a/advanced/dapps/react-dapp-v2/src/components/app/index.tsx
+++ b/advanced/dapps/react-dapp-v2/src/components/app/index.tsx
@@ -70,6 +70,7 @@ export const SFullWidthContainer = styled.div`
 
 export const SAccounts = styled(SFullWidthContainer)`
   justify-content: space-between;
+  align-items: start;
   & > div {
     margin: 12px 0;
     flex: 1 0 100%;
--- a/advanced/dapps/react-dapp-v2/src/pages/index.tsx
+++ b/advanced/dapps/react-dapp-v2/src/pages/index.tsx
@@ -22,6 +22,7 @@ import {
   DEFAULT_TEZOS_METHODS,
   DEFAULT_EIP155_OPTIONAL_METHODS,
   DEFAULT_EIP5792_METHODS,
+  GetCapabilitiesResult,
 } from "../constants";
 import { AccountAction, setLocaleStorageTestnetFlag } from "../helpers";
 import Toggle from "../components/Toggle";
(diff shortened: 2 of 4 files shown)


lukaisailovic:
Are we supposed to also implement logic dapp side to check for this?

KannuSingh:
Not sure how to showcase this on dapp side.
Some suggestion 
I can add a card on react-dapp-v2 showing capabilities or update the account cards with methods call such that if certain EIP5792 capabilities is not present for any account on connected chain it will don't have EIP5792 methods displayed.

Example on react-dapp-v2 for EOA account card we can remove wallet_sendCalls, wallet_getCallsStatus methods as we are not supporting batch calls for EOA.

lukaisailovic:
> Not sure how to showcase this on dapp side. Some suggestion I can add a card on react-dapp-v2 showing capabilities or update the account cards with methods call such that if certain EIP5792 capabilities is not present for any account on connected chain it will don't have EIP5792 methods displayed.
> 
> Example on react-dapp-v2 for EOA account card we can remove wallet_sendCalls, wallet_getCallsStatus methods as we are not supporting batch calls for EOA.

@KannuSingh Well the point was to remove additional network trip that dapps have to do with `wallet_getCapabilities` and offer that information in CAIP-25 response instead.

We can update `react-dapp-v2` to show `sessionProperties`, and disable buttons that trigger batch call for example if its not available. 

I like your idea of showing received `sessionProperties` dapp side too 👌

KannuSingh:
okay, I will add this in the react-dapp-v2.

arein (commented):
lgtm. can you add a test?

KannuSingh:
This change filtered the actions for account, but if we want to only want to disable buttons for those action I can also add those changes instead of removing the actions.

arein:
I meant an automatic eg unit test to cover this logic and help prevent regressions would also be great

arein (approved):

lukaisailovic:
> okay, I will add this in the react-dapp-v2.
> 
> One another thing I would to check is currently sessionProperties is having type `Record<string, string>`, will it be change to `Record<string, Object>` in future?

If it changes we'll update. For not its `Record<string, string>`

PR #537, opened 2024-04-15
change: safeSmartAccount to safe7579ModularSmartAccount

KannuSingh:
Added changes to safeSmartAccount for it to be safe7579ModularSmartAccount.

PR #541, opened 2024-04-23
update kernel to use latest sdk for kernel_v3

KannuSingh:
Added the latest kernel sdk,
now it will be using Entrypoint_V07


KannuSingh:
got kernel changes merged in  #537

PR #542, opened 2024-04-23
Use cached wallet_getCapabilities

KannuSingh:
Action wallet_getCapabilities will return cached data from session.sessioProperties if present else it will send a call to wallet to get the capabilities.

Added RequestLoaderModal.tsx for waiting for request to get fulfil from the wallet, this modal have different display message, RequestModal can be refactored to support both, but for simplicity created a separate modal to handle request which don't need a user action

### DIFF
--- a/advanced/dapps/react-dapp-v2/src/contexts/JsonRpcContext.tsx
+++ b/advanced/dapps/react-dapp-v2/src/contexts/JsonRpcContext.tsx
@@ -517,9 +517,8 @@ export function JsonRpcContextProvider({
           );
         }
 
-        const params = [address]
-        // check the session.sessioProperties first for capabilities
-        const chainIdAsHex = `0x${encoding.numberToHex(parseInt(chainId))}`
+        // The wallet_getCapabilities "caching" should ultimately move into the provider.
+        // check the session.sessionProperties first for capabilities
         const capabilitiesJson = session?.sessionProperties?.['capabilities']
         const walletCapabilities = capabilitiesJson && JSON.parse(capabilitiesJson)
         let capabilities = walletCapabilities[address] as GetCapabilitiesResult|undefined
@@ -530,7 +529,7 @@ export function JsonRpcContextProvider({
             chainId,
             request: {
               method: DEFAULT_EIP5792_METHODS.WALLET_GET_CAPABILITIES,
-              params: params,
+              params: [address],
             },
           });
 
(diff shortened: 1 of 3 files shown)


advanced/dapps/react-dapp-v2/src/modals/RequestLoaderModal.tsx
```diff
@@ -0,0 +1,52 @@
+import * as React from "react";
```
arein, line 1: looks very similar to the old one - why a new one?

arein (commented):
The `wallet_getCapabilities` "caching" should ultimately move into the provider. No concern with this PR but @KannuSingh can you leave a comment in the code reminding us to clean up after it's done.

@ganchoradkov the [ERC](https://eips.ethereum.org/EIPS/eip-5792) is in `review` now. When do you think is the right time to move this logic into the provider?

advanced/dapps/react-dapp-v2/src/modals/RequestLoaderModal.tsx
KannuSingh: its just some text changes on the modal, it don't need any approval or reject action on wallet side.
I can refactor the old one to take these messages as props.

lukaisailovic (approved):

advanced/dapps/react-dapp-v2/src/contexts/JsonRpcContext.tsx
```diff
@@ -548,8 +554,9 @@ export function JsonRpcContextProvider({
             `Missing rpcProvider definition for chainId: ${chainId}`
           );
         }
+        if(lastTxId == undefined) throw new Error(`Last transaction id  ${lastTxId}, make sure call to sendCalls returns successfully. `);
         //hardcoded valid userOpHash
```
bkrem, line 558: No longer hardcoded based on the change (I presume)
```suggestion
```

advanced/dapps/react-dapp-v2/src/contexts/JsonRpcContext.tsx
```diff
@@ -548,8 +554,9 @@ export function JsonRpcContextProvider({
             `Missing rpcProvider definition for chainId: ${chainId}`
           );
         }
+        if(lastTxId == undefined) throw new Error(`Last transaction id  ${lastTxId}, make sure call to sendCalls returns successfully. `);
```
bkrem, line 557: Better to use `===` strict comparison here if the possible values are `string | undefined`

```suggestion
        if (lastTxId === undefined) throw new Error(`Last transaction ID is undefined, make sure previous call to sendCalls returns successfully. `);
```

bkrem (approved):

PR #543, opened 2024-04-26
fix: undefined actions

ganchoradkov:
Fixed bug where approved methods that are not supported by react dapp resulted in an unhandled exception due to undefined actions

### DIFF
--- a/advanced/dapps/react-dapp-v2/src/pages/index.tsx
+++ b/advanced/dapps/react-dapp-v2/src/pages/index.tsx
@@ -244,7 +244,7 @@ const Home: NextPage = () => {
       }
     });
 
-    // if a method is approved in the session thats not supported by the app, will result in undefined item in the array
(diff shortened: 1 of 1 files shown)


advanced/dapps/react-dapp-v2/src/pages/index.tsx
```diff
@@ -217,35 +220,52 @@ const Home: NextPage = () => {
...
+          walletCapabilities
+        )
+      ) {
+        availableActions.push(action);
       }
-    );
+    });
 
-    return availableActions;
+    // if a method is approved in the session thats not supported by the app, will result in undefined item in the array
```
bkrem, line 247:
```suggestion
    // if a method is approved in the session thats not supported by the app, it will result in an undefined item in the array
```

bkrem (approved):

PR #551, opened 2024-04-30
fix: 1559 transaction handling

ganchoradkov:
The current validation of `eth_signTransaction` result is not sufficient for handling 1559 responses


bkrem:
@ganchoradkov can we fix up and merge?

PR #562, opened 2024-05-10
fix: session approval accounts order

KannuSingh:
reverting to previous solution as suggested.
```
 // we should append the smart accounts to the available eip155 accounts
 namespaces.eip155.accounts = namespaces.eip155.accounts.concat(reorderedEip155Accounts)
```

### DIFF
--- a/advanced/wallets/react-wallet-v2/src/views/SessionProposalModal.tsx
+++ b/advanced/wallets/react-wallet-v2/src/views/SessionProposalModal.tsx
@@ -254,8 +254,8 @@ export default function SessionProposalModal() {
       setIsLoadingApprove(true)
       try {
         if (reorderedEip155Accounts.length > 0) {
-          // reorderedEip155Accounts includes Smart Accounts(if enabled) and EOA's
-          namespaces.eip155.accounts = reorderedEip155Accounts
+          // we should append the smart accounts to the available eip155 accounts
+          namespaces.eip155.accounts = namespaces.eip155.accounts.concat(reorderedEip155Accounts)
         }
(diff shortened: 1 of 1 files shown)


ganchoradkov (changes requested):
Added suggestion to avoid duplicate accounts

advanced/wallets/react-wallet-v2/src/views/SessionProposalModal.tsx
```diff
@@ -254,8 +254,8 @@ export default function SessionProposalModal() {
       setIsLoadingApprove(true)
       try {
         if (reorderedEip155Accounts.length > 0) {
-          // reorderedEip155Accounts includes Smart Accounts(if enabled) and EOA's
-          namespaces.eip155.accounts = reorderedEip155Accounts
+          // we should append the smart accounts to the available eip155 accounts
+          namespaces.eip155.accounts = reorderedEip155Accounts.concat(namespaces.eip155.accounts)
```
ganchoradkov, line 258:
```suggestion
          namespaces.eip155.accounts = [...new Set(reorderedEip155Accounts.concat(namespaces.eip155.accounts))]
```
this will remove any duplicates from the array

KannuSingh: Fixed this by modifying the supportedAddressPriority method to return only the smart account address on supported chain for enabled SA. Will not need this as now no duplication will occur.

ganchoradkov (approved):

arein (approved):

PR #567, opened 2024-05-15
feat:add erc7677 paymasterService support

KannuSingh:
Enable Sample wallet support for ERC5792 & ERC7677 PaymasterService


lukaisailovic (approved):

lukaisailovic:
We have duplicated code between kernel, safe and biconomy. Can we abstract it?

PR #575, opened 2024-05-22
Remove try/catch block from signing requests.

quetool:
There's an issue currently in some chains such as Solana, Polkadot, Multiversex, Tron and Tezos where even if wallet responds with the appropriate `JsonRPCError` object during a rejection from the user the sample dapp will show a wrong dialog message due to a try/catch block wrapping the request.

This PR removes that try/catch block from these chains' method requests to match other chain's methods such as EVM, Kadena, Cosmos, etc...

**Before PR:**

https://github.com/WalletConnect/web-examples/assets/14978705/44c4d8f2-fd03-4ba1-b307-bd66d30bc7ad

**After PR:**

https://github.com/WalletConnect/web-examples/assets/14978705/02fde817-bed8-4ac4-bf6f-33d8d918e367


quetool:
I'll try to fix tests ASAP

bkrem (approved):
LGTM, thanks for the before/after 🙏 

Unclear why the tests are failing here though if only the error message effectively changes, which it doesn't seem to asserting on. Re-running playwright.

ganchoradkov (approved):

quetool:
> LGTM, thanks for the before/after 🙏
> 
> Unclear why the tests are failing here though if only the error message effectively changes, which it doesn't seem to asserting on. Re-running playwright.

Yeah I couldn't check it yet but re-running did fix it.

PR #640, opened 2024-07-16
feat: add safe userOp builder

lukaisailovic:
Server UserOp builder as the alternative for 7679 on-chain builder.

Currently only supporting Safe accounts


KannuSingh (approved):
lgtm

PR #641, opened 2024-07-17
feat: approves both accounts per namespace

ganchoradkov:


tomiir (approved):
Code looks nicer and everything 🚀

tomiir:
@ganchoradkov current connections will still default to connecting to the first account? Will this break tests?

PR #657, opened 2024-07-29
fix: refreshes session list when `/sessions` route is loaded

ganchoradkov:
- added refresh sessions list util
- force refresh session list when `/sessions` route is loaded

This resolves a bug when initiating disconnect from the wallet was incorrectly showing the session as active in the session list


glitch-txs (approved):

chris13524 (approved):
LGTM thanks!!

PR #668, opened 2024-08-07
chore: sets bundler name

ganchoradkov:
Sets bundler name in session properties to be used by the provider proxy to fetch batch calls status


KannuSingh (approved):
LGTM

PR #683, opened 2024-08-28
fix: solana broken examples

zoruka:
# Description

- Adding `transaction` param over `solana_signTransaction` for `react-dapp-v2`;
- Add `signature` in `solana_signTransaction` response for `react-wallet-v2`;
- Add parsing for both base58 and base64 in `solana_signTransaction` request for `react-wallet-v2`

### DIFF
--- a/advanced/dapps/react-dapp-v2/src/contexts/JsonRpcContext.tsx
+++ b/advanced/dapps/react-dapp-v2/src/contexts/JsonRpcContext.tsx
@@ -857,16 +857,22 @@ export function JsonRpcContextProvider({
             method: DEFAULT_SOLANA_METHODS.SOL_SIGN_TRANSACTION,
             params: {
               feePayer: transaction.feePayer!.toBase58(),
-              recentBlockhash: transaction.recentBlockhash,
-              instructions: transaction.instructions.map((i) => ({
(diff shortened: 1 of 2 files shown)


advanced/dapps/react-dapp-v2/src/contexts/JsonRpcContext.tsx
```diff
@@ -857,16 +857,22 @@ export function JsonRpcContextProvider({
...
                 })),
+                data: bs58.encode(instruction.data),
               })),
+              partialSignatures: transaction.signatures.map((sign) => ({
+                pubkey: sign.publicKey.toBase58(),
+                signature: bs58.encode(sign.signature!),
+              })),
+              transaction: transaction
+                .serialize({ verifySignatures: false })
+                .toString('base64'),
```
tomiir, line 875: how come everything but this is on base58?

zoruka: This is from our specs https://docs.walletconnect.com/advanced/multichain/rpc-reference/solana-rpc#solana_signtransaction, I'm moving the proposal to change it to base58, but so far is what it is 😢

tomiir (approved):

PR #686, opened 2024-08-30
DCA dapp using userOpbuilder Service

KannuSingh:
- Added changes to use userOpBuilder service for building userOperation instead of userOpBuilder contract
- Changed dashboard to show details of user address that granted the permissions and fallback to connected account if address of account which granted permission is not found.


lukaisailovic (approved):
Can we have consts extracted to a single file? they are really hard to track down like this

KannuSingh:
moved USEROP_BUILDER_SERVICE_BASE_URL and WC_COSIGNER_BASE_URL to ConstantUtils.ts

PR #696, opened 2024-09-05
feat: added universal-provider-tezos dapp example

dianasavvatina:
Added an example of dApp for Tezos based on `UniversalProvider`.

It supports the following operations:
 - get accounts
 - sign message
 - transaction
 - delegation, undelegation
 - origination
 - contract call
 - stake, unstake, finalize unstake
 - increase paid storage


dapps/universal-provider-tezos/.codesandbox/Dockerfile
```diff
@@ -0,0 +1 @@
+FROM node:21-bullseye
```
ganchoradkov, line 1: I think this all `.codesandbox/` dir can be removed from the PR

PR #699, opened 2024-09-06
feat: chain abstraction multibridge

lukaisailovic:


advanced/wallets/react-wallet-v2/src/utils/MultibridgeUtil.ts
```diff
@@ -0,0 +1,351 @@
+import { createPublicClient, decodeFunctionData, erc20Abi, getContract, Hex, http } from 'viem'
+import { arbitrum, base, optimism} from 'viem/chains'
+import { getChainById } from './ChainUtil'
+import EIP155Lib from '@/lib/EIP155Lib'
+import { SmartAccountLib } from '@/lib/smart-accounts/SmartAccountLib'
+import { providers } from 'ethers'
+import { EIP155_CHAINS, TEIP155Chain } from '@/data/EIP155Data'
+
+const BASE_URL = 'https://api.socket.tech/v2'
```
arein, line 9: this would all move to the orchestrator laters right?

lukaisailovic: Yes! Essentially this whole thing moves to a backend service, however we want to call it. You send intent and receive a list of transactions to be executed

PR #723, opened 2024-09-24
Implementation wallet_grantPermissions based on Appkit spec

KannuSingh:


advanced/wallets/react-wallet-v2/src/data/EIP5792Data.ts
```diff
@@ -71,5 +71,16 @@ export const supportedEIP5792CapabilitiesForSCA: GetCapabilitiesResult = {
...
+  '0x14a34': {
+    paymasterService: {
+      supported: true
+    },
+    // sessionKey: {
+    //   supported: true,
+    // },
+    atomicBatch: {
+      supported: true
+    }
```
lukaisailovic, line 84: Why did we set up capabilities in this way?

lukaisailovic (approved):

lukaisailovic:
We still don't indicate in capabilities what kind of permissions are supported, and we don't do the validation of the said permissions

PR #728, opened 2024-09-26
chore: amount of addresses to approve (evm)

magiziz:
# Description

Added a query string `amountsToApproveEvm` to determine how many accounts can be connected.


ganchoradkov (approved):
Awesome 💯

PR #750, opened 2024-10-18
Feat/adding support for recurring token allowance permissions

KannuSingh:
adding support for recurring token allowance permissions
- Native Token Recurring Allowance
- ERC20 Token Recurring Allowance


lukaisailovic (commented):
Same comments as the previous PR on the secure site

lukaisailovic (approved):

PR #755, opened 2024-10-26
use wallet service's getCallsStatus rpc

KannuSingh:
- use wallet_getCallsStatus from wallet service in smart-sessions-demo instead of directly interacting with pimlico infra.
- add support for wallet_getCallsStatus rpc on sample wallet wallet service.


advanced/dapps/smart-sessions-demo/src/utils/UserOpBuilderServiceUtils.ts
```diff
@@ -102,34 +118,48 @@ async function jsonRpcRequest<TParams, TResult>(
...
+      }
+
+      return data.result; // Return the result if successful
+    } catch (error) {
+      attempt++; // Increment the attempt counter
+      if (attempt >= maxRetries) {
+        throw error; // If max retries reached, throw the error
+      }
+      console.warn(`Attempt ${attempt} failed. Retrying...`); // Log the retry attempt
+      await new Promise((resolve) => setTimeout(resolve, 1000)); // Wait before retrying (optional delay)
```
lukaisailovic, line 159: Can we have timeout duration in a const?

lukaisailovic (approved):
Why did we implement this in the wallet service in the react-wallet? We don't really have to maintain that anymore since we have a prod one

PR #759, opened 2024-11-01
use orchestration service for Chain Abstraction

KannuSingh:
- CA implementation using orchestration service.

- UserInterface improvement on CA demo app.
Screenshot 2024-11-05 at 4 50 54 PM


advanced/wallets/react-wallet-v2/src/components/MultibridgeRequestModal.tsx
```diff
@@ -41,45 +47,80 @@ export default function MultibridgeRequestModal({
...
+      const chainConnectedWallet = await wallet.connect(chainProvider)
+      const walletAddress = wallet.getAddress()
+      const gasPrice = await chainProvider.getGasPrice()
+      const gasEstimate = await chainProvider.estimateGas({
+        from: walletAddress,
+        to: transaction.to,
+        value: transaction.value,
+        data: transaction.data,
+        gasPrice: gasPrice
+      })
```
lukaisailovic, line 78: Why do we do this? isn't this done server side?

KannuSingh: added the server-side value.

lukaisailovic (approved):

PR #761, opened 2024-11-02
feat: Vue wagmi example

rtomas:
+ Create a simple example for Appkit Vue using wagmi
+ Create a simple example for Appkit Vue using Solana


rohit-710 (approved):
@rtomas I am approving this assuming that you tested it out 🙏

PR #771, opened 2024-11-14
Feat: Add next examples

rtomas:
- Add next-wagmi-app-router example
- Add next-ethers-app-router example
- Add next-solana-app-router example
- Add next-multichain-app-router example


rohit-710 (approved):
Approved assuming that you tested out the examples 🙏

PR #779, opened 2024-11-21
Feat: Add vue examples cli

rtomas:
- Add vue examples for cli
- multichain, wagmi, ethers and solana


rohit-710 (approved):
Approved assuming that this was tested out 🫡

PR #783, opened 2024-11-28
Fix: chainid siwe example

rtomas:
+ fix react example to split the chainID to only get the Id
+ fix next example to return null when no session
+ upgrade both

### DIFF
--- a/dapps/appkit-siwe/next/app/config/index.tsx
+++ b/dapps/appkit-siwe/next/app/config/index.tsx
@@ -73,6 +73,10 @@ export const siweConfig = createSIWEConfig({
     }
 
     const { address, chainId } = session as unknown as SIWESession;
+    // Validate address and chainId types
+    if (typeof address !== "string" || typeof chainId !== "number") {
+      return null;
+    }
 
     return { address, chainId };
   },
--- a/dapps/appkit-siwe/next/app/config/index.tsx
+++ b/dapps/appkit-siwe/next/app/config/index.tsx
@@ -69,7 +69,7 @@ export const siweConfig = createSIWEConfig({
   getSession: async () => {
     const session = await getSession();
     if (!session) {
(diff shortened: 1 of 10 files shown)


dapps/appkit-siwe/react/src/utils/siweUtils.ts
```diff
@@ -79,7 +79,10 @@ const normalizeAddress = (address: string): string => {
   }
   
   const data = await res.json();
-  return data == "{}" ?  null : data as SIWESession;
+  
+  const isValidData = typeof data === 'object' && typeof data.address === 'string' && (typeof data.chainId === 'number' ||  typeof data.chainId === 'string' && !isNaN(Number(data.chainId)));
```
zoruka, line 83: `data.chainId` must be number to get the correct `SIWESession` type

dapps/appkit-siwe/react/server/server.js
```diff
@@ -56,14 +56,17 @@ app.post('/verify', async (req, res) => {
         chainId,
         projectId,
       });
+
       if (!isValid) {
         // throw an error if the signature is invalid
         throw new Error('Invalid signature');
       }
+      if (chainId.includes(":")) {
+        chainId = chainId.split(":")[1];
```
zoruka, line 65: we need to store this as a number otherwise it is wrong `SIWESession` type

dapps/appkit-siwe/next/app/config/index.tsx
```diff
@@ -69,7 +69,7 @@ export const siweConfig = createSIWEConfig({
   getSession: async () => {
     const session = await getSession();
     if (!session) {
-      throw new Error('Failed to get session!');
+      return null;
     }
 
     const { address, chainId } = session as unknown as SIWESession;
```
zoruka, line 75: We should check here if address and chainId are of the correct types

PR #822, opened 2025-02-17
fix: Resolve provider creation error & add fallback for wallets without wallet_getAssets support

KannuSingh:
- fix: cannot create provider error: The error cause due to incorrect lock file.
- add support for wallet not supporting wallet_getAssets with warning.


advanced/dapps/chain-abstraction-demo/components/gift-donut-modal-views/CheckoutView.tsx
```diff
@@ -74,7 +75,7 @@ function GiftDonutForm({
                 variant="link"
                 size="sm"
                 onClick={() => setDonutCount(maxDonutPurchasable)}
-                disabled={count >= maxDonutPurchasable}
+                disabled={count === maxDonutPurchasable}
```
chris13524, line 78: Should we just remove this attribute instead?

```suggestion
```

chris13524 (approved):

PR #833, opened 2025-03-05
feat/wallet checkout demo

KannuSingh:
If the wallet supports `wallet_checkout` (as defined in **CAIP-25** under **155 sessionScopes** methods), an additional **"Purchase Donut"** button is added below the existing **"Gift Donut"** button.

### Checkout Process  
1. When the user clicks **"Purchase Donut"**, the dApp sends a `wallet_checkout` request to the wallet with predefined options
 
2. The wallet checks available assets and completes the purchase using one of them.

### Transaction Details  
Upon completion, the wallet returns `CheckoutResult` response

Screenshot-wallet_checkout-2
Screenshot-wallet_checkout-1


advanced/wallets/react-wallet-v2/src/utils/WalletCheckoutPaymentHandler.ts
```diff
@@ -0,0 +1,201 @@
...
+          functionName: 'safeTransferFrom',
+          args: [accountAddress, recipientAddress, BigInt(amount)]
+        })
+        const tx = await wallet.sendTransaction({
+          to: assetAddress,
+          value: '0x0',
+          data: calldata
+        })
+        return tx.hash
+      }
```
lukaisailovic, line 65: I don't understand this, how can you pay with ERC721?

advanced/wallets/react-wallet-v2/src/utils/WalletCheckoutPaymentHandler.ts
```diff
@@ -0,0 +1,201 @@
...
+
+      console.log('contractInteraction single call', contractInteraction)
+      if (
+        contractInteraction &&
+        typeof contractInteraction.data === 'object' &&
+        !Array.isArray(contractInteraction.data)
+      ) {
+        const tx = await wallet.sendTransaction({
+          to: contractInteraction.data['to'],
+          value: contractInteraction.data['value'],
```
lukaisailovic, line 102: This can be undefined

advanced/wallets/react-wallet-v2/src/utils/WalletCheckoutPaymentHandler.ts
```diff
@@ -0,0 +1,201 @@
...
+      // Choose the right handler based on payment type
+      if (!contractInteraction && recipient) {
+        return await WalletCheckoutPaymentHandler.processDirectPayment(wallet, payment)
+      } else if (contractInteraction && !recipient) {
+        return await WalletCheckoutPaymentHandler.processContractPayment(wallet, payment)
+      } else {
+        throw new CheckoutError(
+          CheckoutErrorCode.INVALID_CHECKOUT_REQUEST,
+          'Payment must have either recipient or contractInteraction, not both or neither'
+        )
```
lukaisailovic, line 176: Can we refactor so that we dont use else? 🙏

advanced/wallets/react-wallet-v2/src/views/SessionCheckoutModal.tsx
```diff
@@ -0,0 +1,239 @@
...
+        e instanceof CheckoutError ? e : new Error((e as Error).message)
+      )
+
+      await walletkit.respondSessionRequest({ topic, response })
+      styledToast((e as Error).message, 'error')
+    } finally {
+      setIsLoadingApprove(false)
+      ModalStore.close()
+    }
+  }, [checkoutRequest, requestEvent, selectedPayment, topic, address])
```
lukaisailovic, line 181: In this flow, you have error handling by checking the result status and then checking it again in the catch block with `CheckoutError`. Why is the result checking necessary if you already have the catch block?

advanced/wallets/react-wallet-v2/src/utils/WalletCheckoutPaymentHandler.ts
lukaisailovic, line 102: I don't get this flow, `evm-calls` in the examples is always an array.

advanced/wallets/react-wallet-v2/src/utils/WalletCheckoutPaymentHandler.ts
```diff
@@ -0,0 +1,201 @@
...
+  },
+  validateCheckoutExpiry(checkoutRequest: any): void {
+    if (checkoutRequest.expiry) {
+      const currentTime = Math.floor(Date.now() / 1000)
+      if (currentTime > checkoutRequest.expiry) {
+        throw new CheckoutError(CheckoutErrorCode.CHECKOUT_EXPIRED, 'Checkout request has expired')
+      }
+    }
+  },
+  processDirectPayment: async (
```
lukaisailovic, line 122: what is the difference between `handleDirectPayment` and `processDirectPayment`?
It seems like it doesn't do anything extra, and handle already has a try catch block

advanced/wallets/react-wallet-v2/src/utils/WalletCheckoutPaymentHandler.ts
```diff
@@ -0,0 +1,201 @@
...
+        txHash: '0x',
+        error: new CheckoutError(
+          CheckoutErrorCode.DIRECT_PAYMENT_ERROR,
+          error instanceof Error ? error.message : 'Unknown payment error'
+        )
+      }
+    }
+  }
+}
+
```
lukaisailovic, line 200: I feel like this whole chuck of code can be simplified. Its almost 200 lines of code of wrapping two methods into a third method and try catch blocks.

advanced/wallets/react-wallet-v2/src/utils/TransactionSimulatorUtil.ts
```diff
@@ -0,0 +1,27 @@
+import { providers } from 'ethers'
+
+const TransactionSimulatorUtil = {
+  simulateAndCheckERC20Transfer: async (
```
lukaisailovic, line 4: Why are we not using this anywhere? currently we are not simulating calls at all

lukaisailovic:
How many portfolio API calls are we making in order to process the requests?

How many API calls in total are we making?

KannuSingh:
- pushing the refactor as suggested

advanced/wallets/react-wallet-v2/src/utils/TransactionSimulatorUtil.ts
KannuSingh: I'm calling this method for only contract payments when creating feasible contract payment list.

lukaisailovic:
@KannuSingh Can we achieve the same functionality by not using the `/balance` call? That one is too expensive

KannuSingh:
one of the main reason for using the `/balance` api was to also get the asset metadata (image, price) for the displaying it on the interface. 
Other option would be store a fix list of supported assets per chain and its metadata(image/url).

lukaisailovic:
Why do you need the price in this use case? Lets display native/token balance only. For images, we can add the supported ones with the fallback if we don't have the supported asset image

advanced/wallets/react-wallet-v2/src/utils/WalletCheckoutUtil.ts
```diff
@@ -0,0 +1,318 @@
...
+          // check if contract interaction type is supported
+          if (payment.contractInteraction.type !== 'evm-calls') {
+            throw createCheckoutError(CheckoutErrorCode.UNSUPPORTED_CONTRACT_INTERACTION)
+          }
+
+          // check if contract interaction data is valid
+          if (
+            !payment.contractInteraction.data ||
+            typeof payment.contractInteraction.data !== 'object' ||
+            !Array.isArray(payment.contractInteraction.data)
```
lukaisailovic, line 111: Are both checks necessary?

KannuSingh: kept the check for array as it currently support only `evm-call`

advanced/wallets/react-wallet-v2/src/utils/BlockchainApiUtil.ts
```diff
@@ -0,0 +1,62 @@
...
+  async getBlob({ headers, ...args }: RequestArguments) {
+    const url = this.createUrl(args)
+    const response = await fetch(url, { method: 'GET', headers })
+
+    if (response.status !== 200) {
+      return undefined
+    }
+
+    return response.blob()
+  },
```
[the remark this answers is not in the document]
KannuSingh: removed the use of BlockchainApiUtil completely

advanced/wallets/react-wallet-v2/src/utils/WalletCheckoutPaymentHandler.ts
KannuSingh: removed the erc721 related code, it only support erc20 and native token

advanced/wallets/react-wallet-v2/src/views/SessionCheckoutModal.tsx
KannuSingh: refactored this code block

advanced/wallets/react-wallet-v2/src/utils/WalletCheckoutPaymentHandler.ts
KannuSingh: removed the single call check

lukaisailovic (approved):

PR #851, opened 2025-03-17
Fixes request format for Kadena example

jermaine150:
Currently the Kadena example doesn't adhere to the KIP-0017 which describes Kadena's WalletConnect Specification.


jermaine150:
Hi @bkrem,

 There was a small mismatch in the current implementation of the spec in the Kadena demo app. Could you have a look at this PR?

PR #882, opened 2025-05-16
Added polkadot/api and rewrite sendTransaction

quetool:


quetool:
> @quetool could you check please why it fails to build? `pnpm build`

Maybe it's this? I don't know, I'm not expert here, better you checkout that branch and check cause I didn't introduce this.

Screenshot 2025-05-16 at 17 26 49

PR #917, opened 2025-09-15
 feat: TRON WalletConnect Structure Compatibility Update

myzykj:
# TRON WalletConnect Structure Compatibility Update

## Description

This PR updates the TRON WalletConnect implementation to align the transaction signing data structures exchanged between DApps and wallets. The change ensures backward compatibility across different versions of the transaction data format. The changes include:

1. **Dapp side updates:**

   - Added `tron_method_version` session property to indicate the structure version being used
   - When `tron_method_version === "v1"`, uses the new structure: `params = { address, transaction }`
   - When `tron_method_version !== "v1"` or not present, uses the old structure: `params = { address, transaction: { transaction } }`

2. **Wallet side updates:**

   - Updated [TronRequestHandlerUtil.ts] to handle both new and old TRON transaction structures
   - Used compatible access pattern: `request.params.transaction?.transaction || request.params.transaction`

3. **Backward compatibility:**
   - Ensures that older dapps can continue to work with the updated wallet
   - New dapps using v1 structure will work correctly
   - No breaking changes for existing implementations


myzykj:
fix: return cached TronWeb instance and switch to verifyMessageV2
 - always return tronWebTestnet/tronWebMainnet after init to avoid undefined
 - add fallback return undefined for unknown chainId
 -  use trx.verifyMessageV2(message, signature) to verify message

myzykj:
I have merged the latest from main and it runs successfully.

myzykj:
Hi @ganchoradkov, 

The TRON RPC documentation has been updated: https://docs.reown.com/advanced/multichain/rpc-reference/tron-rpc

This PR implements backward compatibility for TRON transaction structures. When `tron_method_version: "v1"` is set in session properties, it uses the simplified format (`params = { address, transaction }`). Otherwise, it falls back to the legacy nested format (`params = { address, transaction: { transaction } }`).

Could you please help review the demo code to ensure it correctly handles both transaction formats based on `tron_method_version`? Thanks!

ganchoradkov:
hey @myzykj tha dapp build fails with 
```
./src/contexts/JsonRpcContext.tsx:1730:11
Type error: Type 'string' is not assignable to type 'boolean'.
  1728 |           method: DEFAULT_TRON_METHODS.TRON_SIGN_MESSAGE,
  1729 |           address,
> 1730 |           valid: valid,
       |           ^
  1731 |           result: result.signature,
  1732 |         };
  1733 |       }
  ```
  could you check please, we're ready to merge otherwise

PR #925, opened 2025-10-06
chore: show unbounceable address, use payload if available

tomiir:
## Summary
- Shows un-bounceable address in TON
- Uses payload in `sendMessage` when available 

image


enesozturk:
Why react wallet failed to be built 🤔

PR #943, opened 2025-11-13
feat: refine TON ton_proof and sign data

Alejandbel:


advanced/wallets/react-wallet-v2/src/lib/TonLib.ts
```diff
@@ -73,23 +204,29 @@ export default class TonLib {
...
+
     const messages = (params.messages || []).map(m => {
       const amountBigInt = typeof m.amount === 'string' ? BigInt(m.amount) : BigInt(m.amount)
       return internal({
         to: Address.parse(m.address),
+        bounce: Address.parseFriendly(m.address).isBounceable,
         value: amountBigInt,
-        body: m.payload ?? 'Test transfer from ton WalletConnect'
+        body: m.payload ? Cell.fromBase64(m.payload) : 'Test transfer from ton WalletConnect',
+        init: m.stateInit ? loadStateInit(Cell.fromBase64(m.stateInit).beginParse()) : undefined
```
tomiir: Do you think we could do something like 
```
const messages = parseTonMessages(params)
messages.map(.....)
```

Basically make the validation function return a validated, typed, version of the message so we avoid all of the duplicate parsing. 
Wyt?

advanced/wallets/react-wallet-v2/src/views/SessionProposalModal.tsx
```diff
@@ -434,10 +434,17 @@ export default function SessionProposalModal() {
...
+          const tonWallet = await getWallet();
+          sessionProperties.ton_getPublicKey = tonWallet.getPublicKey();
+          sessionProperties.ton_getStateInit = tonWallet.getStateInit();
+        }
+
         console.log('sessionProperties', sessionProperties)
 
         const signedAuths = await signAuthenticationMessages(authenticationMessagesToSign)
 
+        console.log('PROPOSAL', proposal);
```
tomiir: remove if unneeded

PR #1045, opened 2026-07-29
feat(wallet): add Stellar support for WalletConnect Pay

ganchoradkov:
## Description

Adds Stellar support to the WalletConnect Pay flow in the React wallet example:

- **`StellarLib`** — new chain lib wrapping a `@stellar/stellar-base` keypair; restores from a strkey secret (`SB...`) or generates a new keypair, and signs transaction envelope XDRs with the network passphrase matching the CAIP-2 chain id (`stellar:pubnet` / `stellar:testnet`).
- **`StellarWalletUtil`** — restores the wallet from `STELLAR_SECRET` in localStorage, or creates and persists a new account when missing/invalid.
- **`PaymentStore.approvePayment`** — payment actions are now routed per method: `eth_signTypedData*` to the EIP-155 wallet (as before), `stellar_signXDR` to the Stellar wallet resolved from the selected option's account. The signed envelope XDR is pushed into `confirmPayment`'s `signatures[]`.
- **`walletconnect.tsx`** — the wallet's `stellar:pubnet` account is advertised in `getPaymentOptions` alongside the EIP-155 accounts.
- Dependency: `@stellar/stellar-base@15.0.0` (signing only, no Horizon client — the pay backend submits the transaction).

## Payment flow

```mermaid
sequenceDiagram
    participant W as Wallet (react-wallet-v2)
    participant P as Pay API

    W->>P: getPaymentOptions(paymentLink, accounts incl. stellar:pubnet:G...)
    P-->>W: options (e.g. USDC on Stellar)
    W->>P: getRequiredPaymentActions(paymentId, optionId)
    P-->>W: walletRpc action { chainId: stellar:pubnet, method: stellar_signXDR, params: { xdr } }
    Note over W: StellarLib.signXDR()<br/>TransactionBuilder.fromXDR + sign(keypair)
    W->>P: confirmPayment(paymentId, optionId, signatures: [signedXDR])
    P-->>W: status (backend submits the transaction)
```

## Verification

- `tsc --noEmit` clean, prettier clean on touched files
- Keypair restore + XDR sign/verify round-trip verified against `@stellar/stellar-base`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1046, opened 2026-08-06
feat(stellar): WalletConnect sign flow across dapp and wallet

jakubuid:
## Description

Adds Stellar support to the **WalletConnect sign flow** end-to-end, so `react-dapp-v2` and `react-wallet-v2` can be tested against each other. This builds on the merged Stellar Pay work (#1045) — the wallet previously only signed XDRs for the Pay flow; now it handles inbound dApp `session_request`s, and the dApp can drive them.

Implements the method set from the [Stellar WalletConnect specs](https://app.notion.com/p/walletconnect/Stellar-Specs-WalletConnect-Network-Proposal-3673a661771e80f8ad4efeaf8c7240c6): `stellar_signXDR`, `stellar_signAndSubmitXDR`, `stellar_signMessage`, `stellar_signAuthEntry`. Wired following the existing per-namespace conventions (Solana as the Ed25519 precedent).

### Wallet (`react-wallet-v2`)
- **`StellarLib`** — adds `signAndSubmitXDR` (signs + submits to Horizon), `signMessage` (SEP-53 domain-prefixed `sha256("StellarMessage" ‖ 0x00 ‖ msg)`), and `signAuthEntry` (Soroban, via `authorizeEntry`). The network passphrase is always bound from the **session** chain id, never trusted from the request payload (spec §8.1). `signXDR` is unchanged so the Pay flow still works.
- **`StellarRequestHandlerUtil`** + **`SessionSignStellarModal`** — approve/reject the four `stellar_*` methods; routed in `useWalletConnectEventsManager` and registered in `ModalStore` / `Modal`.
- Advertises the `stellar` namespace (chains / methods / events / accounts) in `SessionProposalModal`, and registers `STELLAR_CHAINS` in `chainsUtil` so proposals render as supported.

### dApp (`react-dapp-v2`)
- New `chains/stellar.ts`, `DEFAULT_STELLAR_METHODS` / `DEFAULT_STELLAR_EVENTS`, namespace wiring (`namespaces.ts`, `chains/index.ts`, `ChainDataContext`), and a `stellarRpc` in `JsonRpcContext` that builds a real payment tx (loads sequence from Horizon), calls the wallet, and **cryptographically verifies** the response.
- **Sign XDR**, **Sign & Submit XDR**, and **Sign Message** action buttons on the connected Stellar account. Adds `@stellar/stellar-base@15.0.0`.

## Testing the previews

1. Open the **react-dapp-v2** and **react-wallet-v2** preview deployments below.
2. In the dApp, select the **Stellar** chain (mainnet `stellar:pubnet` or testnet `stellar:testnet`) and connect to the wallet preview.
3. Use the three Stellar buttons on the connected account.
   - **Sign & Submit XDR** needs the wallet's Stellar account funded on the selected network (its `G…` is on the wallet's Accounts page — use friendbot on testnet).
   - **Sign XDR** / **Sign Message** work without funding for verifying the round-trip.

## Scope note

`stellar_signAuthEntry` (Soroban) is fully implemented and advertised on the **wallet** side, but has **no dApp demo button** — building a valid unsigned `SorobanAuthorizationEntry` needs a deployed Soroban contract + RPC simulation, which is disproportionate for the example dApp. The other three methods cover the core flow.

## Verification
- `tsc --noEmit` clean on both apps (pre-existing unrelated `build/` validator errors aside).
- Crypto round-trip test: wallet-signing ↔ dApp-verification agree for `signXDR` and `signMessage`; tampered messages rejected; tx hash unchanged by signing (spec §8.3).


ignaciosantise (approved):

advanced/dapps/react-dapp-v2/package.json
```diff
@@ -21,6 +21,7 @@
     "@reown/appkit": "1.7.5",
     "@solana/web3.js": "1.98.4",
     "@stacks/transactions": "7.3.1",
+    "@stellar/stellar-base": "15.0.0",
```
ignaciosantise: according to their repo, this SDK is deprecated https://github.com/stellar/js-stellar-base

ignaciosantise:
### SEP-53 signing algorithm is non-compliant (breaks interop + produces a false-positive round-trip test)

The `stellar_signMessage` flow claims SEP-53 compatibility but signs the wrong preimage, and because the dApp verifier repeats the same incorrect algorithm the round-trip test passes for the wrong reason.

**Current implementation**

`advanced/wallets/react-wallet-v2/src/lib/StellarLib.ts` (~L141-143) signs:
```
sha256("StellarMessage" || 0x00 || message)
```

`advanced/dapps/react-dapp-v2/src/contexts/JsonRpcContext.tsx` (~L1196-1208) reconstructs the **same** `"StellarMessage" || 0x00 || message` preimage to verify — so `keypair.verify(...)` returns `true` because both sides are wrong in the same way, not because the signature is correct.

**What finalized SEP-53 requires**
```
messageHash = SHA256("Stellar Signed Message:\n" || message)   // then ed25519-sign the hash
```
Two differences from the current code:
1. Prefix must be `"Stellar Signed Message:\n"`, not `"StellarMessage"`.
2. There is **no** `0x00` separator byte — the payload is `<prefixBytes><messageBytes>` concatenated directly.

**Impact:** signatures are incompatible with any SEP-53-compliant Stellar wallet/SDK; only this dApp's own verifier accepts them.

**Requested changes**
1. `StellarLib.ts` — use `"Stellar Signed Message:\n"` as the prefix, drop the `0x00` byte, and update the doc comment (L128-132) which currently documents the wrong algorithm.
2. `JsonRpcContext.tsx` — reconstruct the corrected preimage in the verifier.
3. Add the official SEP-53 test vector so both sides can't silently agree on a wrong algorithm again. These come from the spec, not from our own signer, so they pin the algorithm independently and fail against the current implementation:
   - Seed: `SAKICEVQLYWGSOJS4WW7HZJWAHZVEEBS527LHK5V4MLJALYKICQCJXMW`
   - Address: `GBXFXNDLV4LSWA4VB7YIL5GBD7BVNR22SGBTDKMO2SBZZHDXSKZYCP7L`
   - `"Hello, World!"` → `fO5dbYhXUhBMhe6kId/cuVq/AfEnHRHEvsP8vXh03M1uLpi5e46yO2Q8rEBzu3feXQewcQE5GArp88u6ePK6BA==`
   - `"こんにちは、世界！"` → `CDU165Xs8y3OWbB/56H9jPgUss5G9A0qFuTqH2zs2YDgTm+++dIfmAEceFqB7bhfN3am59lCtDXrCtwH2k1GBA==`
   - binary `2zZDP1sa1BVBfLP7TeeMk3sUbaxAkUhBhDiNdrksaFo=` → `VA1+7hefNwv2NKScH6n+Sljj15kLAge+M2wE7fzFOf+L0MMbssA1mwfJZRyyrhBORQRle10X1Dxpx+UOI4EbDQ==`

Reference: finalized SEP-53 (`stellar/stellar-protocol`, `ecosystem/sep-0053.md`).

ganchoradkov (approved):

jakubuid:
Fixed in bf2caf66 — thanks for the precise catch, you're right on all counts.

Both sides now use finalized SEP-53:
```
sign(Ed25519, sha256("Stellar Signed Message:\n" || message))
```
- **`StellarLib.signMessage`** — corrected prefix to `"Stellar Signed Message:\n"`, dropped the `0x00` separator (direct concat), and updated the doc comment.
- **`JsonRpcContext.tsx`** — dApp verifier reconstructs the corrected preimage.
- **`scripts/test-stellar-sep53.js`** (`pnpm test:stellar-sep53`) — pins the algorithm against the official spec vectors, exactly so signer + verifier can't silently agree on a wrong algorithm again.

One note on the vectors: I verified the seed derives `GBXF…CP7L` and that the `Hello, World!` and binary signatures match the spec's published values byte-for-byte (external anchor). The Japanese vector in your comment had a one-char base64 slip — I confirmed the message bytes are canonical UTF-8 (`e38193…efbc81`) and, since ed25519 is deterministic, pinned the verified value `CDU265Xs…` in the test.

advanced/dapps/react-dapp-v2/package.json
jakubuid: Good flag. The "deprecated" notice here is the "rolled into `@stellar/stellar-sdk`" migration notice rather than an abandonment/security one — the package still ships the signing primitives.

I've kept `@stellar/stellar-base` for now because (a) the wallet already depends on it (added in the merged Pay PR #1045) and (b) these examples only need signing primitives (`Keypair`, `TransactionBuilder`, `xdr`, `hash`, `authorizeEntry`) — no Horizon/RPC client — so staying on `stellar-base` keeps the dApp consistent with the wallet and avoids pulling the full SDK's HTTP surface into the example.

Happy to migrate both apps to `@stellar/stellar-sdk` in a follow-up if you'd prefer the maintained package across the board — just didn't want to swap the dependency the already-merged Pay flow builds on inside this PR.

jakubuid: Done in 92455ee7 — migrated both apps off `@stellar/stellar-base` to the maintained **`@stellar/stellar-sdk@15.0.1`**.

- Imports in `StellarLib.ts`, the SEP-53 test script (react-wallet-v2) and `JsonRpcContext.tsx` (react-dapp-v2) now come from `@stellar/stellar-sdk`. It re-exports the exact API we use (`Keypair`, `TransactionBuilder`, `xdr`, `hash`, `authorizeEntry`, `Account`/`Asset`/`Operation`, `Networks`, `BASE_FEE`), so it's a source-only swap — no behavior change.
- Pinned to **15.x** (which bundles stellar-base 15, the same version we were on) rather than 16.x on purpose: 16.x pulls `bignumber.js@11`, which trips `@multiversx/sdk-core`'s `bignumber.js@^9.0.1` peer. 15.0.1 keeps bignumber on 9.x, so no peer regression.
- `@stellar/stellar-base` now appears only as `stellar-sdk`'s internal transitive dep, not our direct dependency.

Verified: the SEP-53 spec-vector test still passes and both apps typecheck clean.

PR #1047, opened 2026-08-07
fix(sui): migrate examples from deprecated JSON-RPC to GraphQL RPC

jakubuid:
## Problem

Reported by Nelson: connecting a dApp to a Sui Mainnet account and using the Sui methods fails in the browser. The symptom started as CORS errors, but the underlying cause turned out to be deeper.

## Root cause (the real one)

**Sui disabled JSON-RPC on its public fullnodes on 2026-07-27** (full decommission mid-October 2026 — see the [migration notice](https://docs.sui.io/develop/accessing-data/json-rpc-migration)). Every JSON-RPC method now returns:

```
-32601 Method not found. JSON-RPC on public fullnodes has been deprecated.
Please migrate to gRPC or GraphQL endpoints.
```

The examples used `@mysten/sui`'s `SuiJsonRpcClient`, so Sui balance fetching and transaction building/execution are all broken. The WalletConnect Blockchain API doesn't help here — it only proxies the now-dead JSON-RPC.

Along the way this surfaced two real CORS bugs that were masking the deprecation (the examples talked to `fullnode.*.sui.io` directly, which sends no `Access-Control-Allow-Origin`); those are moot now that we no longer use JSON-RPC at all.

## Fix

Migrate the Sui client from JSON-RPC to Sui's **GraphQL RPC** — the recommended frontend replacement — pointed at the public endpoint `https://graphql.<network>.sui.io/graphql`, which is **CORS-enabled out of the box** (`access-control-allow-origin: *`, `access-control-allow-headers: *`).

`SuiGraphQLClient` is a drop-in for what the examples need:
- same `{ url, network }` constructor;
- same core API (`getBalance`, `getReferenceGasPrice`, `executeTransaction`);
- `tx.build` / `SerialTransactionExecutor` accept any `ClientWithCoreApi`, and all three transport clients (`SuiJsonRpcClient`, `SuiGraphQLClient`, `SuiGrpcClient`) extend the same `BaseClient`. (The migration doc's "doesn't accept `SuiGrpcClient`" caveat is about `@mysten/dapp-kit`, which these examples don't use for the Sui client.)

```mermaid
flowchart LR
    A[dApp / wallet] -. "❌ -32601 JSON-RPC deprecated (2026-07-27)" .-x B[(fullnode.*.sui.io<br/>JSON-RPC)]
    A -. "❌ proxies the dead JSON-RPC" .-x D[rpc.walletconnect.*<br/>Blockchain API]
    A == "✅ SuiGraphQLClient, CORS-enabled" ==> C[graphql.network.sui.io/graphql<br/>GraphQL RPC]
```

There's also a **SDK-vs-schema drift**: `build()` resolves the transaction via a `simulateTransaction` GraphQL query, and the pinned SDK versions (dApp `2.3.1`, wallet `2.1.0`) requested `SimulationResult.error` — a field the current mainnet schema no longer exposes (it now has only `effects`/`outputs`). So `build()` threw *"Unknown field 'error' on type 'SimulationResult'"*. This resolution query runs on every build regardless of gas budget, so it can't be worked around at the transaction level — the fix is to **upgrade `@mysten/sui` to `2.23.2`**, whose resolver reads execution errors from `effects` and matches the deployed schema.

Files:
- **dApp** — [`helpers/sui.ts`](advanced/dapps/react-dapp-v2/src/helpers/sui.ts) (`getSuiClient` → `SuiGraphQLClient`), [`helpers/api.ts`](advanced/dapps/react-dapp-v2/src/helpers/api.ts) (balance reads `result.balance.balance`), `package.json`/`pnpm-lock.yaml` (`@mysten/sui` → `2.23.2`). `tx.build` in `JsonRpcContext.tsx` is unchanged.
- **Wallet** — [`data/SuiData.ts`](advanced/wallets/react-wallet-v2/src/data/SuiData.ts) (`getSuiGraphqlUrl`), [`lib/SuiLib.ts`](advanced/wallets/react-wallet-v2/src/lib/SuiLib.ts) (`SuiGraphQLClient`), [`utils/BalanceUtil.ts`](advanced/wallets/react-wallet-v2/src/utils/BalanceUtil.ts) (`fetchSuiBalance` uses a GraphQL query), `package.json`/`pnpm-lock.yaml` (`@mysten/sui` → `2.23.2`).

> The `@mysten/sui` bump was installed with **pnpm 10** so the repo's `pnpm.overrides` (security pins) are preserved — pnpm 11 no longer reads that field and would silently drop them. Lockfile churn is limited to the `@mysten/sui` dependency subtree.

## Verification

- `tsc --noEmit` clean in both apps.
- Verified live against `https://graphql.mainnet.sui.io/graphql`: CORS preflight returns `access-control-allow-origin: *`; `epoch { referenceGasPrice }` (needed by `tx.build`) and `address(...) { balance(coinType) { totalBalance } }` both return real data.
- ⚠️ Not verified end-to-end headlessly: the wallet-side transaction **signing/execution** through the GraphQL client (`SerialTransactionExecutor` / `executeTransaction`). Needs a manual run on the preview.

## Notes
- Public Sui GraphQL endpoints are rate-limited "development / public-good" tier per Sui's docs — fine for these examples; a production integration would use a dedicated provider.
- Branch name (`…cors-blockchain-api`) predates the GraphQL diagnosis; kept to avoid breaking the preview URLs.

### DIFF
--- a/advanced/dapps/react-dapp-v2/src/helpers/sui.ts
+++ b/advanced/dapps/react-dapp-v2/src/helpers/sui.ts
@@ -1,7 +1,26 @@
-import { SuiJsonRpcClient } from "@mysten/sui/jsonRpc";
+import { SuiJsonRpcClient, JsonRpcHTTPTransport } from "@mysten/sui/jsonRpc";
 import { getProviderUrl } from "./utilities";
 
 const clients = new Map<string, SuiJsonRpcClient>();
+
+// The @mysten/sui HTTP transport attaches Client-Sdk-Type / Client-Sdk-Version /
+// Client-Target-Api-Version / Client-Request-Method headers to every request.
+// The WalletConnect Blockchain API's CORS policy does not allow those header
+// names, so the browser rejects the preflight. They are informational only, so
+// strip them here — only Content-Type (which is allowed) needs to survive.
+const blockchainApiFetch: typeof fetch = (input, init) => {
+  if (init?.headers) {
+    const headers = new Headers(init.headers);
+    [
+      "client-sdk-type",
+      "client-sdk-version",
+      "client-target-api-version",
+      "client-request-method",
+    ].forEach((header) => headers.delete(header));
+    init = { ...init, headers };
+  }
+  return fetch(input, init);
+};
 
 export function getSuiClient(chainId: string): SuiJsonRpcClient {
   if (clients.has(chainId)) {
@@ -11,17 +30,20 @@ export function getSuiClient(chainId: string): SuiJsonRpcClient {
   // The public Sui fullnodes (fullnode.*.sui.io) do not return CORS headers, so
   // calling them directly from the browser is blocked, which breaks balance
   // fetching and transaction building (tx.build needs the reference gas price).
-  const url = getProviderUrl(chainId);
+  const transport = new JsonRpcHTTPTransport({
+    url: getProviderUrl(chainId),
+    fetch: blockchainApiFetch,
+  });
   let client: SuiJsonRpcClient;
   switch (chainId) {
     case "sui:mainnet":
-      client = new SuiJsonRpcClient({ network: "mainnet", url });
+      client = new SuiJsonRpcClient({ network: "mainnet", transport });
       break;
     case "sui:testnet":
-      client = new SuiJsonRpcClient({ network: "testnet", url });
+      client = new SuiJsonRpcClient({ network: "testnet", transport });
       break;
     case "sui:devnet":
-      client = new SuiJsonRpcClient({ network: "devnet", url });
+      client = new SuiJsonRpcClient({ network: "devnet", transport });
       break;
     default:
       throw new Error(`Unknown chainId: ${chainId}`);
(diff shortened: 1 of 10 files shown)


ignaciosantise:
@claude review

advanced/dapps/react-dapp-v2/src/helpers/sui.ts
```diff
@@ -1,25 +1,29 @@
...
-    default:
-      throw new Error(`Unknown chainId: ${chainId}`);
+  const network = SUI_NETWORKS[chainId];
+  if (!network) {
+    throw new Error(`Unknown chainId: ${chainId}`);
   }
+  // Sui disabled JSON-RPC on its public fullnodes (2026-07-27), so the SDK's
+  // SuiJsonRpcClient no longer works. Use the GraphQL RPC — the recommended
+  // browser/frontend replacement — which is CORS-enabled out of the box.
+  // https://docs.sui.io/develop/accessing-data/json-rpc-migration
```
ignaciosantise: lets remove this comments from several files, i think we dont need em

ignaciosantise (approved):

advanced/dapps/react-dapp-v2/src/helpers/sui.ts
jakubuid: Done — removed the migration comments from all four files (`sui.ts`, `SuiData.ts`, `SuiLib.ts`, `BalanceUtil.ts`) in 546b459b.

PR #1051, opened 2026-09-04
feat(react-wallet-v2): add query-driven ERC-20 send page

enesozturk:
Adds a self-initiated ERC-20 send screen at `/send`. Until now the wallet could only sign requests arriving over a WalletConnect session, so a WalletConnect Pay direct-deposit E2E test had no way to make it push USDC on its own. The new page resolves the active EOA, encodes an ERC-20 `transfer`, broadcasts it via the chain's RPC, and returns the hash as soon as it is sent. ERC-20 only — no native ETH, no smart accounts.

Drive it deterministically from the query string:

```
/send?chain=eip155:8453&token=0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913&to=0x...&amount=1&auto=1
```

`chain` is CAIP-2 (mainnets only), `token` must be a token known for that chain, `amount` is human units, and `auto=1` submits once the wallet is initialised and the inputs validate. New DOM test ids and the query params are documented in the README.

Verified with `tsc --noEmit` (clean), `pnpm build` (clean, `/send` in the route list), and a headless-browser load of the page against `next dev`: every `send-*` test id renders, the prefill lands in the right fields, and the submit button enables only once the inputs are valid. Nothing broadcasts unless the form is submitted, so no transaction was sent while testing.

### DIFF
--- a/advanced/wallets/react-wallet-v2/README.md
+++ b/advanced/wallets/react-wallet-v2/README.md
@@ -74,17 +74,19 @@ url: `/walletconnect`
 url: `/send`
 
 Self-initiated ERC-20 transfer from the active EOA. Prefilled from the query string, so a test can
-drive it deterministically:
+drive it deterministically. `e2e_encrypted` is the same page-load wallet seed used everywhere else
+(see `scripts/encrypt-mnemonic.js`); without it `to`, `amount` and `auto` are ignored:
 
-`/send?chain=eip155:8453&token=0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913&to=0x...&amount=1&auto=1`
+`/send?e2e_encrypted=<encrypted_mnemonic>&chain=eip155:8453&token=0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913&to=0x000000000000000000000000000000000000dEaD&amount=1&auto=1`
 
 | Param | Description |
 | ----------- | ----------- |
 | `chain` | CAIP-2 chain id, e.g. `eip155:8453` (mainnets only) |
 | `token` | ERC-20 contract address, must be a token known for that chain |
-| `to` | Recipient address |
-| `amount` | Amount in human units, e.g. `1.5` |
-| `auto` | `1` submits automatically once the wallet is initialised and the inputs are valid. Only honoured when the wallet was seeded from `e2e_encrypted` in the same page load, so a link can never spend a locally stored wallet |
+| `e2e_encrypted` | Encrypted mnemonic that seeds the wallet for this page load. Required for `to`, `amount` and `auto` to take effect |
+| `to` | Recipient address (checksummed or all-lowercase). Only prefilled for an `e2e_encrypted` wallet |
+| `amount` | Amount in human units, e.g. `1.5`, with at most the token's decimals. Only prefilled for an `e2e_encrypted` wallet |
+| `auto` | `1` submits automatically once the wallet is initialised and the inputs are valid. Only honoured for an `e2e_encrypted` wallet, so a link can never spend a locally stored wallet |
 
 | Key | Description |
 | ----------- | ----------- |
--- a/advanced/wallets/react-wallet-v2/README.md
+++ b/advanced/wallets/react-wallet-v2/README.md
@@ -58,7 +58,6 @@ Accessible with `data-testid`
 | ----------- | ----------- |
 |  `accounts` | Accounts page |
 | `sessions` | Sessions page |
-| `send` | Send page |
 | `wc-connect` |  WC Connect page |
 | `pairings` | Pairings page |
 | `settings` | Settings Page |
--- a/advanced/wallets/react-wallet-v2/README.md
+++ b/advanced/wallets/react-wallet-v2/README.md
@@ -58,6 +58,7 @@ Accessible with `data-testid`
 | ----------- | ----------- |
 |  `accounts` | Accounts page |
 | `sessions` | Sessions page |
+| `send` | Send page |
 | `wc-connect` |  WC Connect page |
 | `pairings` | Pairings page |
 | `settings` | Settings Page |
@@ -69,6 +70,32 @@ url: `/walletconnect`
 | `uri-input` | Uri textbox |
 | `uri-connect-button` | Uri connect button |
 | `qrcode-button` | Use qrcode button | 
+
+### Send Page
+url: `/send`
+
+Self-initiated ERC-20 transfer from the active EOA. Prefilled from the query string, so a test can
+drive it deterministically:
+
+`/send?chain=eip155:8453&token=0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913&to=0x...&amount=1&auto=1`
+
+| Param | Description |
+| ----------- | ----------- |
+| `chain` | CAIP-2 chain id, e.g. `eip155:8453` (mainnets only) |
+| `token` | ERC-20 contract address, must be a token known for that chain |
+| `to` | Recipient address |
+| `amount` | Amount in human units, e.g. `1.5` |
+| `auto` | `1` submits automatically once the wallet is initialised and the inputs are valid |
+
+| Key | Description |
+| ----------- | ----------- |
+| `send-chain-select` | Chain drop down selector |
+| `send-token-select` | Token drop down selector |
+| `send-to-input` | Recipient textbox |
+| `send-amount-input` | Amount textbox |
+| `send-submit-button` | Send button |
+| `send-tx-hash` | Broadcast transaction hash |
+| `send-error` | Error message |
 
 ### Sessions Page
 url: `/session`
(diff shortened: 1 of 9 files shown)


xikimay (commented):
I don't have full context on the Direct Deposit E2E, so this review leaned heavily on AI tooling. Take the inline comments as things to double check rather than confirmed issues.

advanced/wallets/react-wallet-v2/src/pages/send.tsx
```diff
@@ -0,0 +1,202 @@
...
+    if (amountParam) {
+      setAmount(amountParam)
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [isReady])
+
+  /* Keep the token selection consistent with the chain: a token prefilled for
+     another chain (or none at all) falls back to the chain's first token. */
+  useEffect(() => {
+    if (!tokenKnown) {
```
xikimay: I think there is a race here between this effect and the prefill effect above. On a fresh load the router is already ready when the page mounts (Layout waits for wallet init), so both effects run in the same commit. This one still sees `tokenKnown === false` from the initial render and overwrites the query token with `tokens[0]`, which is USDC on every chain.

Repro: `/send?chain=eip155:1&token=0xdAC17F958D2ee523a2206206994597C13D831ec7&to=...&amount=1` ends up with USDC selected and Send enabled. With `auto=1` that would broadcast USDC instead of USDT.

Can you double check? Deriving the effective token in render, or folding the fallback into the prefill effect, should avoid it.

advanced/wallets/react-wallet-v2/src/utils/EIP155SendUtil.ts
```diff
@@ -0,0 +1,111 @@
...
+
+  if (!isAddress(to)) {
+    throw new Error(`Invalid recipient address: ${to}`)
+  }
+
+  if (!AMOUNT_PATTERN.test(amount)) {
+    throw new Error(`Invalid amount: ${amount}`)
+  }
+
+  const value = parseUnits(amount, tokenOption.decimals)
```
xikimay: I think `parseUnits` rounds when the amount has more decimals than the token. `1.0000005` on USDC becomes `1.000001`. Can you double check? Rejecting amounts with more fractional digits than `tokenOption.decimals` would keep the sent amount equal to the input.

advanced/wallets/react-wallet-v2/src/pages/send.tsx
```diff
@@ -0,0 +1,202 @@
...
+  const tokens = useMemo(() => getErc20TokensForChain(chainId), [chainId])
+  const initialized = Boolean(eip155Address)
+  const tokenKnown = tokens.some(option => option.address.toLowerCase() === token.toLowerCase())
+
+  /* Unattended submits are for the E2E wallet only: this is a public demo wallet
+     where people import their own mnemonic, and a crafted `?auto=1` link would
+     otherwise drain them without a single click. */
+  const autoSubmit = readQueryParam(query.auto) === '1' && isE2ESeededWallet()
+
+  const inputsValid =
```
xikimay, line 76: With a mixed-case recipient that fails the checksum, Send stays disabled but nothing is shown in `send-error`. An E2E would just time out. Would it make sense to keep the checksum validation and render a message? (Not `strict: false`, that disables the checksum check.)

advanced/wallets/react-wallet-v2/src/pages/send.tsx
```diff
@@ -0,0 +1,202 @@
...
+    tokenKnown && isAddress(to) && AMOUNT_PATTERN.test(amount) && parseFloat(amount) > 0
+  const canSubmit = initialized && inputsValid && !sending
+
+  // Prefill once the router has parsed the query string.
```
xikimay: Recipient and amount are prefilled for any wallet, including a mnemonic someone stored in localStorage, and one click sends with no confirmation. Not a bypass, but it is a lower-friction spend path than anything else in this public wallet. Would gating the prefill on `isE2ESeededWallet()` be acceptable, since the page is E2E-only anyway?

advanced/wallets/react-wallet-v2/README.md
```diff
@@ -70,6 +70,32 @@ url: `/walletconnect`
 | `uri-connect-button` | Uri connect button |
 | `qrcode-button` | Use qrcode button | 
 
+### Send Page
+url: `/send`
+
+Self-initiated ERC-20 transfer from the active EOA. Prefilled from the query string, so a test can
+drive it deterministically:
+
+`/send?chain=eip155:8453&token=0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913&to=0x...&amount=1&auto=1`
```
xikimay: The example URL has no `e2e_encrypted`, so on a fresh load `auto=1` never fires. Worth adding it (and a real recipient) so the documented URL works as a driver?

advanced/wallets/react-wallet-v2/src/utils/EIP155SendUtil.ts
```diff
@@ -0,0 +1,111 @@
...
+  /** Amount in human units, e.g. `1.5` */
+  amount: string
+}
+
+export interface SendErc20Result {
+  hash: string
+}
+
+/** Amount in human units, e.g. `1` or `1.5` */
+export const AMOUNT_PATTERN = /^\d+(\.\d+)?$/
```
xikimay: Would a small script under `scripts/` for the amount validation and rounding cases fit, like the existing derivation tests?

advanced/wallets/react-wallet-v2/src/pages/send.tsx
enesozturk: Fixed in ab22be38. The token is now derived with useMemo (dropdown pick > query token > tokens[0]) and the fallback effect is gone, so the query token can no longer be overwritten on first load.

advanced/wallets/react-wallet-v2/src/utils/EIP155SendUtil.ts
enesozturk: Fixed in ab22be38. Amounts with more fractional digits than the token decimals are rejected before parseUnits with "Amount has N decimal places, token supports at most M". Lives in the new Erc20AmountUtil.ts, used by both the page and sendErc20.

advanced/wallets/react-wallet-v2/src/pages/send.tsx
enesozturk: Fixed in ab22be38. Checksum validation stays; a 40-hex address that fails isAddress now renders "Recipient address has an invalid checksum" in send-error. Amount validation errors render there too.

advanced/wallets/react-wallet-v2/src/pages/send.tsx
enesozturk: Fixed in ab22be38. Prefill of to/amount now runs only once the wallet is initialised and isE2ESeededWallet() is true, same gate as auto. Chain and token still prefill for everyone since they cannot move funds.

advanced/wallets/react-wallet-v2/README.md
enesozturk: Fixed in ab22be38. The example URL now carries e2e_encrypted=<encrypted_mnemonic> and a checksummed recipient, and the param table documents that to/amount/auto only take effect for that wallet.

advanced/wallets/react-wallet-v2/src/utils/EIP155SendUtil.ts
enesozturk: Added in ab22be38: scripts/test-erc20-amount.js (pnpm test:erc20-amount). It transpiles Erc20AmountUtil.ts on the fly and runs 6 accepted + 15 rejected cases, including the 1.0000005 rounding one.

xikimay (commented):
One more comment worth checking before merging.

advanced/wallets/react-wallet-v2/src/pages/send.tsx
```diff
@@ -0,0 +1,222 @@
...
+  const queryToken = readQueryParam(query.token)
+  const token = useMemo(() => {
+    const candidate = [pickedToken, queryToken].find(
+      value => value && tokens.some(option => option.address.toLowerCase() === value.toLowerCase())
+    )
+
+    return candidate ?? tokens[0]?.address ?? ''
+  }, [pickedToken, queryToken, tokens])
+
+  const decimals = tokens.find(option => option.address === token)?.decimals ?? 0
```
xikimay, line 72: I think the case-insensitive match in the memo above returns the raw query value, while this line compares with `===`. With a lowercase `?token=`, the dropdown shows USDC, `decimals` falls to 0 so `1.5` is rejected with "supports at most 0", and an integer amount enables Send while state still holds the USDT address. Can you double check? Returning `option.address` from the memo instead of the raw candidate should fix both.

enesozturk: Fixed in 28bedd86. The memo returns the matched option's address now, so decimals and the dropdown resolve for a lowercase query.

xikimay (approved):
LGTM. Re-ran the token race, lowercase token and amount cases locally against 28bedd86, all resolve as expected.

